PR #168, opened 2015-01-12
Added spline kernel (from P. Hennig) to GPy

tjhgit:
Had to modify the base_plots and model_plots.py, since I had troubles
installing GPy in anaconda on debian linux due to the dependency on
Tango. Why is Tango needed to represent colors that can also be typed in
Hex format thus eliminating further dependencies.

### DIFF
--- a/GPy/kern/__init__.py
+++ b/GPy/kern/__init__.py
@@ -13,6 +13,7 @@ from _src.ODE_UYC import ODE_UYC
 from _src.ODE_st import ODE_st
 from _src.ODE_t import ODE_t
 from _src.poly import Poly
+from _src.spline import Spline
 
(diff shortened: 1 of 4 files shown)


tjhgit:
Hi James,

updated points 1) and 2) in the latest commit. I do not know how to do point 3.

Regards,
Thomas.

jameshensman:
Hi @tjhgit , 

Thanks for the edits, this looks nice. 

What's the domain of the kernel -- is it valid for any input X? I think previous spline kernels I've seen have been defined on [0,1]. Same here?

Would it be possible to implement update_gradients_diag? This would allow use with the sparse GPs in GPy. 

Thanks, 

James.

PR #174, opened 2015-02-08
Kernel product bugfix

beckdaniel:
A bugfix to prevent a zero division in kernel products when one or more kernel evaluations result in zero. Added two tests: one for update_gradients_full and another one for gradients_X.


jameshensman:
Thanks Daniel, looks good to me.

PR #176, opened 2015-02-16
Fixed Bernouli Likelihood for correct predictive variance on GP Classification tasks

iassael:


jameshensman:
Hi @iassael, 

Thanks for the fix! 

James.

PR #180, opened 2015-02-26
Devel

mikecroucher:
Baby steps towards Python 3 compatibility while retaining Python 2 compatibility. Please use this PR instead of #179 since this also fixes most exceptions for Py3 compatibility and resolves a conflict from upstream.


mikecroucher:
Too many things are broken for this to be merged into the main devel branch

PR #183, opened 2015-03-10
ENH: Adding standard periodic kernel

AlexGrig:
The standard periodic kernel is the one which is mentioned
in the Rasmussen and Williams book about Gaussian Processes.


jameshensman:
HI @AlexGrig , 

This code looks good, thanks for contributing!

There have been a couple of requests for the periodic kernel. I know @alansaul and @asolin have been looking at it. Yesterday, Alan and I were discussing how  this might be done using the stationary kernel, which does lots of the heavy-lifting for you. In particular, it would allow gradients_X, with no extra work required from you, as well as allowing much shorter code for update_gradients_full. 

Perhaps we could amalgamate your code here with the stationary proposals?

Cheers,

James.

AlexGrig:
Hi, James

I am in touch with @asolin regarding this (we are from the same university). I can look at stationary kernel class in more detail to use it for periodic kernel. I think it is possible to do since periodic kernel is a function of r.

Alex.

AlexGrig:
Hi, James

I have looked more closely to the stationary kernel.
I see a problem here of making use of stationary kernel for this standard periodic kernel. Current implementation of stationary kernel is actually isotropic (or homogeneous) kernel, where kernel parameter "r" is a scalar. I looked at formulas and I think that if the stationary kernel were dependent on vector "r" (r= x - x') then it would have been possible to utilize it. In the current situation I do not know how to do it. Maybe you and Alan have something in mind?


Regards,
Alex.

Issue #199, opened 2015-06-03
Kernel Tutorial Needs updating


lawrennd:
This page needs updating to the new style of kernel objects (or removing and an appropriate link to a relevant notebook). 

https://gpy.readthedocs.org/en/latest/tuto_kernel_overview.html?highlight=kernel%20tutorial

mzwiessele:
this is weird, i thought this has been done already? i remember on one hackday we had a distribution of doc updates, i did the model part and someone else said he or she would do the kernel part, does that ring a bell for someone?

lawrennd:
Can you pop up a link to how this material is accessed in case someone is
willing but unable due to not knowing (like I was in a brief window when I
saw this problem).

Neil

mzwiessele:
ne just needs to update the corresponding file in GPy/doc/<something i dont remember>/tuto_kernel_overview.rst

the build is being done automatically on commit, thanks to @asaul!

jameshensman:
We need to ensure these changes get carried over to 0.8 (They should, when devel is merged onto master)

Issue #192, opened 2015-05-01
Tensor Parameter does not work


antonywaldock:
When using the simple example below
kern1 = GPy.kern.RBF(1) + GPy.kern.White(1)
kern2 = GPy.kern.RBF(1)
kernx = kern1.prod(kern2,tensor=True)

but I get the error

TypeError: prod() got an unexpected keyword argument 'tensor'

tensor does not appear to be a valid option. The code used to work in an older version (July 2014)

jameshensman:
Hi @antonywaldock , 

The interface for this has changed between releases. Sorry if it caused you trouble. the correct way to implement this would be

kern1 = GPy.kern.RBF(1, active_dims=[0]) + GPy.kern.White(1)
kern2 = GPy.kern.RBF(1, active_dims=[1])
kernx = kern1 * kern2

(note that the active dimensions of the white kernel are irrelevant)

PR #208, opened 2015-07-19
Tutorial and readme updates

strongh:
This is mostly an attempt at #199. Also some minor tweaks for docstrings and the README.

The only part I'm not sure about is what happened to the tensor param for the `add` and `prod` methods, #192. I've removed mentions of the param from the docstrings and tutorial to reflect the current implementation.


jameshensman:
Awesome, many thanks.

PR #210, opened 2015-08-05
Warped GP mods/additions

beckdaniel:
Additions:
- A method to estimate the mean with Gauss-Hermite quadrature.
- A method to estimate the variance, also with GH.
- A method to calculate quantiles in the warped (observation) space.
- Identity warping function: mainly for tests and sanity checks.
- A test that compares a standard GP with a Warped GP using the Identity warping function (they must give the same results).

Modifications:
- Previously, warpedGPs gave medians as predictions. I changed the default behaviour to output the mean instead. The median can still be obtained by passing "median=True" to predict.
- The way the variance was calculated before was by just passing the inverse warping function to the latent gaussian variance, which I assumed it was wrong. Now it uses the GH method.
- Default behaviour for predictions was to predict in the latent space. I changed that to predict in the warped space as default instead (flag predict_in_warped_space defaults to True).
- Plotting now plots everything in the warped space as default (median + quantiles). Latent space plots can be retrieved by setting "predict_in_warped_space" to False before plotting.

Comments:
- There is also a test (skipped by default) in which I tried to replicate Snelson's et al. original toy 1D dataset (cubic root of a noisy sine function). However, I couldn't replicate the experiment because the Newton-Raphson method in f_inv seems to be diverging in many values inside the grid. The resulting plot gives nonsense values for the median and the quantiles in these places. I currently have no idea how to make NR behave properly in this setting... In the same test there's code to try np.exp(np.sin(X)), which seems to work fine, unlike Snelson's example.


jameshensman:
Hi @beckdaniel , 

This looks great, thanks!

Could you just have a little tidy in the warped_gp model class, please? there are still a few dev-style comments that could be tidied away (like #MEDIAN!)

@nfusi , do you have anything to add?

beckdaniel:
Hi @jameshensman 

Sorry about the delay in replying. Which dev-style comments do you actually mean? The "#MEDIAN!" one, for example, was deleted on commit 20a1e8d (the second to last one).

I am also not sure if this Travis error is due to something I changed in the code. The log seems to be complaining about the lack of cblas.h somewhere.

mzwiessele:
This looks really good, indeed. Can you rebase so we see if the tests pass? We are working*very* hard on having the test coverage up there in order to make sure to have the user run updates without hesitating. Thus, when your tests pass (after re basing to the current devel) we will merge the request happily : )

Thank you for the (very good) effort!

beckdaniel:
I can do it, yes. Any reasons for rebasing instead of merging? (I only ask because I never used rebase before, so I'm more comfortable with merge)

mzwiessele:
No merging is the way to go, sry for the confusion!

beckdaniel:
Done, merged the last devel. Travis CI seems to be still running so I'm not sure if all the tests are passing. Running nosetests gives me one error on "likelihood_tests.py" and one intermitent failure on "rv_transformation_tests.py", which seem unrelated to my changes (but I might be wrong).

mzwiessele:
The tests seem to fail, can you check what the problem is?

mzwiessele:
Oh I looked at it, it looks good! I will try to merge it in, thank yo so much!

PR #211, opened 2015-08-10
Fixes the PDF transformation bug by handpicking James Hensman's code from the devel branch

ebilionis:
This patch addresses the following bugs:
1) The PDF computed by the Model class does not factor in the determinant of the Jacobian of the transformation of the hyper-parameters.

2) If you try to plot a transformation you get an exception.

Bug number 1, was already fixed by James Hensman. I just handpicked his code from the devel branch and wrote a test case (see GPy/testing/rv_transformation_tests.py). Note that the test case fails for the current branch and succeeds only after this change is made. Uncomment, the matplotlib related code in the test case to see the problem with your own eyes.

Bug number 2 was trivial to fix.


ebilionis:
BTW, the failure is a standard failure in the master branch.

lawrennd:
@zhenwendai can you have a quick check through @ebilionis  code and merge in the pull request?

Thanks Ilias!

zhenwendai:
Thanks Ilias for the bug fix! It is not merged into the master branch.

Issue #205, opened 2015-07-14
[devel] missing "cblas.h"


zhenwendai:
The current devel branch needs "cblas.h". Cross-platform support needs to be properly solved in future.

A quick fix for Ubuntu:
apt-get install libblas-dev

nfusi:
Yes, this doesn't work on Windows and doesn't take advantage of MKL on Linux and Mac. Scipy.linalg.blas seems to have all the required functions (ddot,  dsymv, dscal). @jameshensman do you know if there's anything that prevents a cython rewrite? Looks like you were working on cython choleskies at the time of the commit.

alansaul:
The reason for the cblas import was to make a super quick c implementation which is wrapped up with Cython, but the c implementation relies on a BLAS routine (about 10x faster than the pure Cython implementation we had previously I believe). I hadn't realised that it doesn't use MKL though? I thought the BLAS routines would use them by default but your probably right!

We should definitely make it fall back if cblas it can't be found, as for the real bottlenecks it would be nice to be able to make C + BLAS + MKL implementations, but we shouldn't make BLAS a dependency for obvious reasons. @mikecroucher has said that he will look at this when he gets a minute. In cholesky_backprop.c there is a pure c version for the routine in question chol_backprop / old_chol_backprop.

nfusi:
Great! I _think_ that you can call scipy.linalg.blas functions from cython code, so that you don't have to give up the BLAS (and MKL) speedup when going from C to cython.

alansaul:
Cool! We'll look into that thanks Nicolo! Hope you're well.

mathDR:
I just posted an issue resolving this on Yosemite.  The vecLib header files were moved on 10.10.4

jameshensman:
the problem with calling the scipy.linalg.blas routins from cython is they don;t release the gil, so you['ll get almost no speedup over a python implementation. 

In the upcoming scipy release (16? 17?), there are cython bindings for lapack and blas, which will release the gil. So the fix is either:

 1) wait for the next scipy, re-write the reverse-model cholesky differentiation in cython (cf cython + c)
 2) work out how to compile against a local blas cleanly

@mikecroucher is looking at (2).

Dapid:
They are in Scipy 0.16 in `scipy.linalg.cython_blas` and `scipy.linalg.cython_lapack`. It was released the 24th of July.

Here are the wrapped functions:
[...]

PR #214, opened 2015-08-17
ENH: implementing the Cholesky backpropagation through Scipy's BLAS

Dapid:
Here it is. I have also added a few enhancements to the rest of the file that slipped into the same commit.

Closes https://github.com/SheffieldML/GPy/issues/205.


Dapid:
All tests pass in my local installation (Fedora 21, Python 2.7.8, NumPy 1.9.2, SciPy 0.16, Cython 0.23).

jameshensman:
Many many thanks for this. I merged this PR manually after a few additional commits  -- it turned out that my tests didn't cover this case!

the code is now available on devel.

Dapid:
Great!

I have seen there are more Cholesky related functions, mostly in utils/linalg.py, that are a mixture of BLAS, numpy, ctypes, and weave. Is there any plan to clean this up? Specially now we are depending on Scipy's Cython interface to BLAS we can replace most of the weave and ctypes implementations.

Doing a simple GPRegression on master, almost 50% of the time is spent on that module (23% on dpotri, 15% on jitchol, and 10% in dtrti). Of these, jitchol looks like it may benefit from Cython.

PR #215, opened 2015-08-17
Update travis

Dapid:
This is an update to Travis' new framework. I am also ditching conda and using plain pip.

Python 3 added to track the compatibility progress.


mzwiessele:
As we require scipy 0.16 for the new release, this might not work anymore?

PR #224, opened 2015-09-03
Fix print statement

Dapid:
I noticed there is a small incompatibility between Python 2 and 3, as it is using `print` with kwargs, which is a syntax error in Python 2.


mikecroucher:
thanks

Dapid:
Since this was not caught earlier, this file is not being tested. In fact, I have just realised, it is importing two modules that are not declared in setup.py nor used. Fixing it in: https://github.com/SheffieldML/GPy/pull/227

mikecroucher:
Test coverage is much lower than it should be. We need to do better.

PR #225, opened 2015-09-03
Cross Platform Cython Fixes

mikecroucher:
I don't think that the mechanism we were attempting to use to detect a lack of Cython-built modules was working properly so I've changed it. There is some tidying up left to do and a few cross-platform issues have been exposed (I'll open the issues later) but this at least allows the test suite to be run even when Cython modules are not built.

If the Cython modules are not built, all functions fall back to the pure Python or Python+numpy implementations. The Cython-specific tests are skipped.

On Ubuntu, on a fresh git clone:

nosetests ./GPy/testing/
SSSSSSS......... /home/walkingrandomly/GPy/GPy/core/parameterization/transformations.py:77: RuntimeWarning:invalid value encountered in greater
 /home/walkingrandomly/GPy/GPy/core/parameterization/transformations.py:82: RuntimeWarning:invalid value encountered in greater
........................................................................................................................................................................................................................................S.S...............................................................
----------------------------------------------------------------------
Ran 314 tests in 12.828s

OK (SKIP=9)

Now, I build Cython modules:

    python setup.py build_ext --inplace

nosetests ./GPy/testing/................ /home/walkingrandomly/GPy/GPy/core/parameterization/transformations.py:77: RuntimeWarning:invalid value encountered in greater
 /home/walkingrandomly/GPy/GPy/core/parameterization/transformations.py:82: RuntimeWarning:invalid value encountered in greater
........................................................................................................................................................................................................................................S.S...............................................................
----------------------------------------------------------------------
Ran 314 tests in 12.340s

OK (SKIP=2)

I deleted a couple of Cython related tests from the kernel testing module. This is because it seemed tricky to propagate the relevant variable through the full calculation without modifying the interface to kernel functions. As such, I may have decreased Cython test coverage (need to check!) but the equivalence of Cython/Python implementations should really be checked more directly anyway IMHO.

I'm not going to merge this change myself since I'd prefer someone else to review and merge (or not).

Cheers,
Mike

Update - Mac OS X Cython support
 I've added fixes so that Cython modules now compile on Mac OS X. Setup.py detects if it's running on a Mac and, if it is, removes references to OpenMP. A similar mechanism is in the .c include header. 

This will be a pain for users of GNU gcc on Mac since they won't get OpenMP even though they have OpenMP but, if they REALLY need it, they can always manually edit the relevant files for now. 

I propose not to worry about this too much since Clang will get OpenMP eventually and we can just remove this stuff.

### DIFF
--- a/GPy/kern/_src/coregionalize.py
+++ b/GPy/kern/_src/coregionalize.py
@@ -5,7 +5,6 @@ from .kern import Kern
 import numpy as np
 from ...core.parameterization import Param
 from ...core.parameterization.transformations import Logexp
-from ...util.config import config # for assesing whether to use cython
 try:
     from . import coregionalize_cython
     cython_available = True
@@ -84,7 +83,7 @@ class Coregionalize(Kern):
     def Kdiag(self, X):
         return np.diag(self.B)[np.asarray(X, dtype=np.int).flatten()]
 
-    def update_gradients_full(self, dL_dK, X, X2=None):
+    def update_gradients_full(self, dL_dK, X, X2=None,cython=False):
         index = np.asarray(X, dtype=np.int)
         if X2 is None:
             index2 = index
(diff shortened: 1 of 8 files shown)


mikecroucher:
@jameshensman Would you mind giving this the once over please? If accepted, we can run on all 3 platforms without Cython.

The next step will be build instructions for all 3 platforms and bug squashing since it seems that there are some hidden cross-platform bugs!

PR #230, opened 2015-09-08
Update travis

Dapid:
Rebasing #215  to be compatible.

Installing GPy with plain pip and testing both Python2 and 3.


Dapid:
Scipy takes a bit more than 10 min on Travis to install, and if there is no output, it gets killed; so I am running pip in verbose mode. Subsequent runs should run immediately, because it is cached.

Some tests are failing, though. One is `test_inferenceX_BGPLVM` in `self.assertTrue(np.allclose(m.X.mean, mi.X.mean, rtol=1e-4, atol=1e-4))`, and the other is the missing overflow warning.

mzwiessele:
This is great, sorry for being late to the party : )

Can you rebase to devel, so I can merge it in?  We put six into the dependencies as well, that seems to be the conflict?

mzwiessele:
I have created another pull request from dapids changes. Thank you so much @Dapid this is great and really appreciated!!!

This is the change which should be from @Dapid: #246

PR #248, opened 2015-09-12
Travis update to test Linux & MacOSX

mzwiessele:
Added compatibility for OSX Travis testing using Miniconda


mzwiessele:
What is this? I have added in some more tests, so that 20 checks get done. Also added in a random seed. Can you please have a look at this build?
https://travis-ci.org/SheffieldML/GPy/builds/80271579

The numerical gradients are different between versions (weird, as there is a seed), the gradients sometimes check, and sometimes do not and on some versions the gradients are plain out wrong. What is going on there?

                                         !TRICKERY!

PR #261, opened 2015-10-08
Kalman new

AlexGrig:
State-Space inference for 1 dimensional Gaussian Processes.


mu2013:
Has AlexGrig change anything?  I am working on the version from 19
september 2015. It works fine. although the plot  need to be fixed.  I add
spatiol temporal state space GP. But currently the gradient is working in
temporal direction.

mu

lawrennd:
HI Mu,

Alex was with us at the last summer school, he's working with Simo and has
been developing the state space approach for covariance functions for
integration. I'm hoping we can integrate this pull request as quick as
possible, if you want to check changes please do go through them and
comment!

Neil

AlexGrig:
Hi, Mu,
So you are working on something similar - spatio-temporal KF inference? Could you given some reference of what exactly you are doing? Maybe you could use some parts of my code, and I can provide comments by e-mail.

By the way, some tests are not passed for this PR. I suppose this is because random number generator produce different values on a test machine and on my computer. Also, some my tests may take a relatively long time, because they involve hyper-parameter optimization. I still think think that they are useful, but what is your opinion should I do anything about it?

mu2013:
Hi Alex
Thanks for your reply. I have just seen some of your changes. The old
version only have sde version of kernel for matern32 and ou.  You have
included all kinds of kernels which can be transformed to sde. That's
perfect!  I have also realised the framwork of kalamn_new branch has been
changed. you use some cython code which I do not understand.

I am trying to make a spatiol temporal GP but using SDE in temporal
direction. In spatiol direction we can using a normal GP with any type of
kernel.  At the moment the spatio-temporal kernel is working, but I only
make the gradient of likelihood working in temporal direction. But all of
this work is till using the old framework develped by arno solin in Mar
2015. I do not know if it is still possible for me to run the
spatio-temporal GP after I pull your changes. In the worst case, I can have
a separate copy of the branch.

But if you are also interested in sptiao-temporal GP or high dimension
input GP, maybe we can work together and sort this out? I learned the link
of GP and SDE from simo and arno. are you simo's posdoc or phd?

very nice work!

AlexGrig:
Hi, Mu

Thanks for reply. Yes, I have changed the framework, we discussed that with Arno. The main idea is to untie the state-space code from the GPy itself. Now there is basically one point when GPy model calls the state-space inference - in the file state_space_model. So, the code became modular, and for you, I suppose, it should also be easier to use it. 

Regarding the Cython, the idea was to speed up the inference especially for optimizing hyper parameters, because I have relatively long time series. However, currently Cython provides almost no speedup. Maybe I implemented it improperly or maybe the cycles are not the bottleneck, rather linear algebra is. Anyway, I decided to still keep in repository, but its usage is unjustified. There is a parameters "use_cython" which you can play with yourself.

Ok, I'll send you an e-mail in order to keep this Pull Request discussion and other stuff separately.

Regards,
Alex.

Dapid:
@AlexGrig have you profiled the code? That will show you exactly in which functions time is being spent. Compile Cython with the directive #cython profile=True to include hooks to profile inside the functions.

I see a few possible improvements in the Cython code, I'll add them.

GPy/models/state_space_cython.pyx
```diff
@@ -0,0 +1,964 @@
+# -*- coding: utf-8 -*-
+"""
+Contains some cython code for state space modelling.
+"""
+import numpy as np
+cimport numpy as np
+import scipy as sp
```
Dapid, line 7: The recommended way to import scipy is `from scipy import linalg`.

GPy/models/state_space_cython.pyx
```diff
@@ -0,0 +1,964 @@
...
+
+    cpdef reset(self,compute_derivatives = False):
+        raise NotImplemented("(cython) reset is not implemented!")
+
+cdef class R_handling_Cython(Measurement_Callables_Cython):
+    """
+    The calss handles noise matrix R.
+    """
+    cdef:    
+        np.ndarray R
```
Dapid, line 71: Typing as ndarray helps very little without adding number of dimensions and, specially, dtype. Once that is in place, you can add `boundscheck=False` for faster indexing, that will make a big difference.

GPy/models/state_space_cython.pyx
```diff
@@ -0,0 +1,964 @@
...
+        
+        cdef np.ndarray[DTYPE_t, ndim=2] inv_square_root
+        
+        cdef np.ndarray[DTYPE_t, ndim=2] U
+        cdef np.ndarray[DTYPE_t, ndim=1] S           
+        cdef np.ndarray[DTYPE_t, ndim=2] Vh
+        
+        if (R.shape[0] == 1): # 1-D case handle simplier. No storage
+        # of the result, just compute it each time.
+            inv_square_root = np.sqrt( 1.0/R )
```
Dapid, line 143: If this was taking a significant amount of time, it is faster to iterate over the elements.

GPy/models/state_space_cython.pyx
```diff
@@ -0,0 +1,964 @@
...
+        else:
+            if self.svd_each_time:
+                
+                U,S,Vh = sp.linalg.svd( R,full_matrices=False, compute_uv=True,
```
Dapid, line 147: Here you can use the Cython interface to linalg. I used it in `GPy/util/choleskies_cython.pyx`, for reference.

GPy/models/state_space_cython.pyx
```diff
@@ -0,0 +1,964 @@
...
+                
+                inv_square_root = U * 1.0/np.sqrt(S)
+            else:
+                if ind in self.R_square_root:
+                    inv_square_root = self.R_square_root[ind]
+                else:
+                    U,S,Vh = sp.linalg.svd( R,full_matrices=False, compute_uv=True, 
+                              overwrite_a=False,check_finite=True)
+                              
+                    inv_square_root = U * 1.0/np.sqrt(S)
```
Dapid, line 158: Same as before, iterate over elements instead of going through Numpy's indexing.

GPy/models/state_space_cython.pyx
```diff
@@ -0,0 +1,964 @@
...
+                    
+                    self.R_square_root[ind] = inv_square_root
+                
+        return inv_square_root
```
Dapid, line 162: Once you have typed what you are missing, and use the Cython API for linalg, I think you will be able to define the function as `nogil` and gain a bit extra. Some temporaries will still be required, but these can be private class members created in `__init__`.

GPy/models/state_space_cython.pyx
```diff
@@ -0,0 +1,964 @@
...
+cimport cython
+
+#from libc.math cimport isnan # for nan checking in kalman filter cycle
+cdef extern from "numpy/npy_math.h":
+    bint npy_isnan(double x)
+
+DTYPE = np.float64
+DTYPE_int = np.int64
+
+ctypedef np.float64_t DTYPE_t
```
Dapid, line 17: This would be a good case for fused types. `cython.floating` stands for float and doubles (important when you want to save some memory).

GPy/models/state_space_cython.pyx
```diff
@@ -0,0 +1,964 @@
...
+ctypedef np.int64_t DTYPE_int_t
```
Dapid, line 18: You are using this for indexing; being pedanting, it should be `Py_ssize_t` to avoid overflow.

GPy/models/state_space_cython.pyx
```diff
@@ -0,0 +1,964 @@
...
+        
+        cdef np.ndarray[DTYPE_t, ndim=2] square_root
+        
+        cdef np.ndarray[DTYPE_t, ndim=2] U
+        cdef np.ndarray[DTYPE_t, ndim=1] S           
+        cdef np.ndarray[DTYPE_t, ndim=2] Vh
+        
+        if (Q.shape[0] == 1): # 1-D case handle simplier. No storage
+        # of the result, just compute it each time.
+            square_root = np.sqrt( Q )
```
Dapid, line 296: Here `square_root` is untyped. You should initialise it first in the cdef and pass the values:
`square_root[:] =`

or just `np.sqrt(Q, out=square_root)`.

This applies in many places.

For temporary arrays that will always have the same shape, you may get better performance allocating it at `__init__` time and reusing them.

GPy/models/state_space_cython.pyx
```diff
@@ -0,0 +1,964 @@
...
+    cdef np.ndarray[DTYPE_t, ndim=2] dQ
+    if calc_grad_log_likelihood:
+        dA_all_params = p_dynamic_callables.dAk(k) # derivatives of A wrt parameters 
+        dQ_all_params = p_dynamic_callables.dQk(k) # derivatives of Q wrt parameters
+        
+        param_number = p_dP.shape[2]
+        
+        # p_dm, p_dP - derivatives form the previoius step
+        dm_pred = np.empty((p_dm.shape[0], p_dm.shape[1], p_dm.shape[2]), dtype = DTYPE)
+        dP_pred = np.empty((p_dP.shape[0], p_dP.shape[1], p_dP.shape[2]), dtype = DTYPE)
```
Dapid, line 625: You are overriding the typing and making the GC do extra work. Allocate memory once, and reuse.

Dapid:
If you provide me with a benchmark and a clean set of passing tests, I can help improving the performance in the Cython code, but I think my comments (and `cython -a `) should go a long way.

AlexGrig:
Hi, Dapid

Thanks for your comments, I have used the profile only for Python code not Cython. I will compose a benchmark for Python and Cython to test the speed difference. I will also try to take your advises into account.

Regarding the tests, probably, first I need to check them on Python 3.5 to see what cause errors. I also hope that if I set the seed of a random number generator explicitly, then then random numbers are exactly the same on all operating systems and python versions. If this is not the case then tests must be designed for each configuration. You can run test on your machine, maybe they work well.

Regards,
Alex.

Dapid:
The random numbers, given the seed, should be the same in all platforms and versions; otherwise it is a serious bug. A while back there was some discussion in the NumPy list about it, and the consensus seems to be that it is a very important feature to dedicate effort to keep.

Regarding profiling: if you add the directive:

    # cython: profile=True

you can already profile it under cProfile. I'd do it myself, but I'll need a relevant test case.

The problem with Travis seems to be that some test is taking too long, perhaps you can split a big test into many smaller ones? Each test should take much less than 10 min on the (possibly puny) machines Travis has for testing, or it will get killed. I'll see what happens on my machine.

lawrennd:
Good stuff everyone!

Just for info we are trying to ensure that for any cython code there is
also a python fall back incase we can't compile (it sounds like that's
happening here already!).

Neil

AlexGrig:
@Dapid  Hi, Dapid, I will prepare the test case soon, just a bit busy at the moment.
@lawrennd Hi, Neil, yes that is right the python code is used when the cython is not available, but I did not know that this is a policy ), now I know.

Dapid:
On 13 October 2015 at 12:03, AlexGrig <notifications@github.com> wrote:

> @Dapid <https://github.com/Dapid> Hi, Dapid, I will prepare the test case
> soon, just a bit busy at the moment.


Don't hurry, I have a deadline coming up soon too.

mzwiessele:
@AlexGrig great work so far, it would be great to have this in the new GPy release. Could we get an update on the tests, so they actually check?

mu2013:
I had a conversation with Alex and have read through part of his code. But
the plotting  is not working in his branch at the moment.

cheers
mu

AlexGrig:
Hello, Guys

Yes, I will try to bring this branch into shape including tests and
plotting.
Max could you say when is the planned release date? I think I ll start
working tomorrow.

Regards,
    Alex Grigorevskiy.

mzwiessele:
The new release shall get out as soon as possible!

We are extremely overdue and is going very slowely…

mzwiessele:
This looks great, now some of the new code is not fully covered. We want all new code to be covered by the tests, there is already enough code not being run during testing time. We are working hard to get the coverage up from all the other modules, but progress is slow.

Here you can see your coverage:
https://codecov.io/github/SheffieldML/GPy/GPy?ref=7e1dd02771fa317813d3ed36c15bcd2a5a339baf

Thanks

mzwiessele:
Thanks Alex really great work!

PR #268, opened 2015-10-15
Improvements to the binomial likelihood and GP_var_gauss

jameshensman:
GP_var Gauss can now use likelihoods which require metadata. 

The binomial likelihood (with metadata) is now more numerically stable.


mzwiessele:
Thank you James, that's great!

We will merge it in as soon as we have the automised binary wheel creation for distribution of the package installed. 

The next milestone has this in!

mzwiessele:
fixed

PR #269, opened 2015-10-15
Paramz

mzwiessele:
pulled out paramz and made sure everything works fine in GPy

testing paramz now and making sure it is properly stable.


alansaul:
Will this mean that everyone who has ever written a GPy model or component
will have to rewrite it or else it will no longer work?! Seems like quite a
big deal!!

Alan

zhenwendai:
In general, it is a good direction to separate parameterization framework out of GPy. However, a lot of our extensions rely on the current setting of GPy. We should at least leave it there with some warnings of deprecation for a reasonable long time (e.g. a year) before cutting it out of GPy.

mzwiessele:
Nope it will work just fine.

mzwiessele:
But doing the deprecation is a good thing!

I will go through the relevant bits in the imports and do some deprecations.

It should be done in a way so that it will not be problamatic though.

mzwiessele:
Actually I am trying to do it, so that the average developer doesnt even know, that it has been pulled out.
So lets see, maybe you want to try it out as soon as the tests pass.

PR #279, opened 2015-11-20
Fixing likelihoods and EP

alansaul:
This is a pull request to fix some problems with the Bernoulli and EP I have noticed recently.

First of all I think the signs of the Bernoulli seemed messed up, and the tests hid this problem.

```
In [2]: import numpy as np

In [3]: A = (np.random.randn(5,1) > 0.0)*1.0

In [4]: A
Out[4]:
array([[ 0.],
       [ 0.],
       [ 0.],
       [ 1.],
       [ 0.]])

In [5]: A = A*2 - 1

In [6]: A
Out[6]:
array([[-1.],
       [-1.],
       [-1.],
       [ 1.],
       [-1.]])

In [7]: np.where(A)
Out[7]: (array([0, 1, 2, 3, 4]), array([0, 0, 0, 0, 0]))

In [8]: np.where(A == 1)
Out[8]: (array([3]), array([0]))
```

We were previously doing np.where(y, this, elsethat), where I think we should be doing np.where(y==1, this, elsethat).

I have added a setting to EP which automatically resets after every step (though the default behaviour is the same), as the interface is currently confusing for doing true EP updates. I think this could be further resolved, and we could definitely do with looking at how we can speed this up, I believe the EP method could be significantly faster than it is at the moment.

The EP marginal likelihood reported is currently incorrect, this is because there is a missing term contributing to the marginal likelihood (the third and fourth term in the GPML book, p57). This won't change the optimization (as these terms are constant in the hyper parameters), but will report the marginal likelihood correctly. I have independently checked the marginal likelihoods with another GP package (GPStuff) and they appear to be correct.

I haven't been able to check this for the EPDTC code as of yet, but I believe the contributing terms should be similar? @ric70x7 can you verify this as I am not super familiar with the sparse approximation for EP? Maybe we could add an additional test where we compare the marginal likelihood with EPDTC vs EP, and check they are the same when Z=X?

Alan


ric70x7:
Hi,
I'm sorry for the delay, I was planning to make a notebook showing how the
approximations match with the right parameters, but I just haven't got the
time.
The contributing terms for EP-DTC and EP should be similar. The comparisons
I have done so far return "similar" likelihoods.
About the test, we would just need to be careful because we might get some
numerical differences. Then we would need some criteria to define that both
cases "match".
Thanks a lot for spending the time and going through the trouble of fixing
this Alan.
Cheers,
Ricardo

2015-11-20 14:46 GMT+00:00 Max Zwiessele notifications@github.com:

PR #280, opened 2015-11-20
Update MANIFEST.in

mzwiessele:


Dapid:
The `.so` are still included.

PR #281, opened 2015-11-24
Addition of Kronecker inference for Grid Inputs + Gauss-Markov Models for 1-D Inputs

kurtCutajar:
This work extends GPy with an implementation of Gauss-Markov models for single-dimensional data, and fast Kronecker inference for grid inputs. The implementation of these two models is largely stand-alone within GPy and should not interfere with the existing functionality.

Both approaches enable inference in linear time for the indicated structured inputs. This work is based on some of the models (and their corresponding implementations) discussed in the PhD thesis 'Scalable inference for structured Gaussian process models' (2012), by Yunus Saatçi, and 'Scaling multidimensional inference for structured gaussian processes' (2015), by Cunningham et al.

This work has been carried out under the supervision of Dr. Maurizio Filippone, who has previously also collaborated on extending GPy. Further implementation details may be found below.

---

**Kronecker Inference for Grid Inputs**

GPy already contains a _GPKroneckerGaussianRegression_ model for carrying out fast inference on two-dimensional grid inputs. However, due to the fact that the implementation is hard-coded for two dimensions, thus rendering it inflexible, we decided to implement the procedure from scratch as a separate model, _GpGrid_, and having a wrapper called _GPRegressionGrid_.

This class inherits most of its functionality from the standard GP model, but also redefines the following procedures:
- _parameters_changed_ - In the standard case, the derivatives may be computed using the information contained in the GP posterior returned from the inference procedure. In this case, the evaluated derivatives are returned by the inference function itself, thus requiring no further computations;
- _raw_predict_ - The procedure for making predictions at new input locations also requires updating. To achieve this goal, we implement a new function, kron_mmprod, which extends upon the kron_mvprod procedure. This simply amounts to executing the latter function M times, where M denotes the number of test inputs;

The computation of covariance for this algorithm relies on breaking down a regular kernel function into a tensor product over a scalar function evaluated for every dimension. In order to achieve this functionality, we introduce a _getOneDimensionalKernel_ method, which returns the scalar variant of the designated kernel. This scalar kernel must in turn implement the methods defined by a _GridKerns_ class, which is intended to be a superclass for all such scalar kernels. Thus far, we have only provided an implementation for the RBF kernel, which can be decomposed into the scalar product of the _GridRBF_ kernel evaluated over the individual dimensions. Finally, we also implemented a _GridPosterior_ object, which is returned upon completion of the inference procedure, and can then be used for making predictions over test data, etc.

This model can be invoked in a manner similar to that of the standard GP regression model. An example for 5-dimensional data is provided below:

_kernel = GPy.kern.RBF(input_dim=5, variance=1, ARD=True)_
_gridModel = GPy.models.GPRegressionGrid(X, Y, kernel)_

---

**Gauss-Markov Model for 1D inputs**

We introduced a _GpSSM_ class which implements a Gauss-Markov model for one-dimensional inputs. Once again, the _GpSSM_ model extends the regular GP class whilst overriding certain methods, as follows:
- _optimize_ - Unlike the regular GP model, the optimisation is not carried out by directly maximising the marginal likelihood. Instead, we were required to implement an Expectation-Maximisation procedure. For the maximisation step, we decided to use the Limited-memory BFGS algorithm, which is one of the most widely-used optimisation algorithms for hyperparameter learning;
- _parameters_changed_ - Given that we do not compute derivatives of the likelihood w.r.t. the parameters in this model, this step was omitted in our implementation, and we simply carry out inference using the updated parameters when this method is invoked;
- _raw_predict_ - The procedure for making predictions is markedly different in this model, since it relies solely on the conditional distributions at the neighbouring states (input locations) in the Markov chain representation of the GP;

We also introduced a wrapper for the _GpSSM_ class, _GPRegressionSSM_, so that the former model is indirectly instantiated through this class. As expected, we also introduced a _GaussianSSMInference_ class, which contains the main implementation for constructing the Gauss-Markov model assuming Gaussian likelihood. The inference procedure yields _SsmPosterior_, which is expressly intended for the SSM in order to represent a Gaussian posterior over latent function values. This contains attributes which are necessary for both the optimisation and prediction procedures. Finally, we also introduced a new class for kernel functions which are suitable for Gauss-Markov models, _SsmKerns_, which in turn inherits methods from the base kernel class. This class mandates that all sub-classing kernel functions implement a series of necessary procedures, such as the transition functions from one state to another, as well as computing derivatives of the transition and covariance functions w.r.t. the parameters of the model. So far, we have only adapted the Matérn-3/2 kernel for use with this model. However, more kernels can easily be added provided that they implement all of the required functionality.

As in the previous case, this model can be accessed by invoking the indicated wrapper class, and setting the necessary parameters of the kernel:

_kernel = GPy.kern.Matern32_SSM(input_dim=1, variance=1, lengthscale=1)_
_ssmModel = GPy.models.GPRegressionSSM(X, Y, kernel)￼_


Dapid:
This is very cool!

I haven't reviewed the code, but I have two comments: first, you seem to have deleted a bunch of docs, I imagine it was accidental, so it should be reverted.

The second is regarding style, `getOneDimensionalKernel` is a method, so it should be called `get_one_dimensional_kernel`, or some shorter bikeshedded variant. CamelCase in Python is reserved for class names.

lawrennd:
Excellent! Firstly great news @kurtCutajar that you've done this, and
thanks so much @Dapid for looking at any issues in the pull request.
@mzwiessele do you think we can integrate rapidly given the fixes @Dapid is
suggesting?

mzwiessele:
This is great indeed! 

It would be greatly beneficial if you could do the suggested pep adjustments:
- _pep8_ https://www.python.org/dev/peps/pep-0008/
- _pep20_ https://www.python.org/dev/peps/pep-0020/

Also, and for me more importantly, you could try to go over to the devel branch, as the test coverage has been greatly increased over there. Also is there gradient checks to be done? I understand you are using an EM-like optimisation, so I am not sure.

What is about the docs? I saw that as well, and was wondering why you deleted that part? On the devel branch the documentation has changed significantly, so this might help with getting the docs back into this branch.

We will actually put the devel branch release out next week, so if you'd like this to be in the next release, you need to do it before next week. 

Thanks so much, this is great work!

kurtCutajar:
Thank you all for your feedback!

I deeply apologise for the late reply, but I was caught up with some other work commitments and missed the notifications for this conversation thread!

Removing the documentation was indeed a mistake on my part - apologies! I will ensure that this change is reverted. The style recommendations have also been noted! This was my first time working in Python so I'll be a bit more careful in the future. I will also follow your suggestion to merge my changes with the development branch first too.

I plan on having a go at the recommended fixes over the December break, and I will get back to you in due course. Hope this works well for you!

Best regards,
Kurt

mzwiessele:
I hope this is not forgotten? Could you merge in the current devel branch, so we can see the checks run again? 

Also, can you go over to the devel branch, instead of master? 
https://help.github.com/articles/using-pull-requests/#changing-the-branch-range-and-destination-repository

mzwiessele:
bump

kurtCutajar:
Apologies once again - for some reason I never get any email updates for this pull request.

I had a chance to look at the most recent developments and noted that some notable changes have been effected, particularly the separation of the Parameterisation functions.  In view of these modifications, I decided to re-fork the GPy repo as I figured it would be easier to re-introduce my code from scratch, and indeed had no problem with the GpGrid code.

However, I encountered a somewhat more pertinent issue for the SSM model which relies on a two-step EM procedure which was originally also slightly troublesome to introduce (I believe GPy doesn't yet support such forms of optimisation explicitly). I had originally included this in the code by way of a not-so-elegant hack, but a new approach is required as a result the afore-mentioned separation of concerns. 

I haven't yet figured out which is the best way to go about this problem, but hope to (very soon) set aside some time to do it properly.

mzwiessele:
Sounds great! I am aware, that GPy does not allow for EM like optimization, as it is meant for gradient based opitmization. The best way of implementing an EM approach in my opinion would be to put a new optimizer in, which knows about EM. This would then need an additional function (E-step is parameters_changed()) for the M-step or something like it. The default optimizer for your method would then be the EM optimizer, which can then be reused by other people who need an EM optimizer?

kurtCutajar:
I will give it a go as soon as possible. I apologise once again for the constant delays.

PR #285, opened 2015-12-08
Warped GPs improvements

beckdaniel:
Hi all,

These are some small updates on Warped GPs.
- The Newton-Raphson procedure was initializing the values of f^-1(z) with +1. This was causing divergence problems when z is negative. Now it initializes y with +1 where z is positive and -1 where it is negative.
- Solving the divergence issues let me replicate Snelson's original toy example (a sine function pushed through a cubic root). There's a test that implements it and plots a bunch of stuff:
  - A normal GP
  - A Warped GP in warped space
  - A Warped GP in latent space
  - The warping function used in the Warped GP.
    The test has a skip decorator because since it plots stuff it halts the automated testing procedure. Comment the decorator to see the plots. (A better solution on this is appreciated).
- Changed some stuff on the plotting functions to make sure Warped GPs are plotted in the right way, including data.


mzwiessele:
Could you merge devel into your branch, so that the tests run again?

beckdaniel:
Will do but it might take a few days (paper deadline mode).

beckdaniel:
Hi, it took less than expected...

There are a few changes compared to the first pull request:
- Added a learning rate parameter to the warping function, which is used when calculating the inverse using NR. This is performing better than the initialization trick I described before so I took it out.
- The numeric inverse is now part of the WarpingFunction mother class. If an inherited function has a closed form solution it can just override the f_inv method.
- Added a log warping function.
- There were two tanh implementations in the code, I merged them into a single one and renamed it to TanhFunction.
- Tests for the Identity and Log function now explicitly use the numeric inverse, even though they have closed formulas.

mzwiessele:
Great, thank you so much @beckdaniel ! I will merge it in, do we have some example, which can be run in order to show the good work you did?

beckdaniel:
I had a test on model_tests.py which replicated the original 1-D cubic sine function from Snelson et al. paper. But now I see that the examples folder is probably a better place for it so I moved it there.

lawrennd:
Thanks @beckdaniel does it remain as a test? I.e. was it a copy rather than
a move :-)

beckdaniel:
No, it was never a test in the first place since it lacks any actual assertions, it just train two models and plot the results.

lawrennd:
If you have time to make a quick test on the code, that would be _really
great_. Helps us keep coverage up! It's best to try and test the
sub-components of the model (like the specific warping code).

mzwiessele:
That would be _great_ indeed! This will also ensure, that your code will be usable also in the future : )

beckdaniel:
Sure, didn't realize I was not covering TanhFunction in the current tests. I will come up with some more tests but that might take a while.

lawrennd:
Thanks Daniel!

mzwiessele:
Thanks Daniel, that's great work : )

mzwiessele:
Thank you so much, at risk of repeating myself, great work daniel!

Now one last thing, the warping functions are not being tested fully, check

https://codecov.io/github/SheffieldML/GPy/GPy/util/warping_functions.py?ref=6296a2127b43e480f9aaa6d685dc5915c35b7094

Can we get the warping functions module to 100% coverage? It does not have to be very sophisticated tests, but it is good to have the code run fully during test time, so no changes will be overseen to effect other code, when pushing (or making a pull request).

mzwiessele:
Bump

GPy/plotting/gpy_plot/plot_util.py
```diff
@@ -31,6 +31,7 @@
 import numpy as np
 from scipy import sparse
 import itertools
+from GPy.models import WarpedGP
```
mzwiessele:
``` python
from ...models import WarpedGP
```

GPy/models/warped_gp.py
```diff
@@ -5,21 +5,23 @@
 from ..util.warping_functions import *
 from ..core import GP
 from .. import likelihoods
-from GPy.util.warping_functions import TanhWarpingFunction_d
+from paramz import ObsAr
+from GPy.util.warping_functions import TanhFunction
```
mzwiessele:
``` python
from ..util.warping_functions import TanhFunction
```

mzwiessele:
Great work so far, lets get it ready for GPy : )

There is still coverage missing. Can you cover the transformations fully with your tests (also branching points) so we can get these great features in?

beckdaniel:
Hi, sorry about the delay in this. I am addressing some of the coverage stuff now

mzwiessele:
no worries, i just dont want it to get forgotten :) it is some nice features you are implementing!

beckdaniel:
I replicated Snelson's cubic root example into a test. There's no assertions on it, it is there to ensure coverage on the tanh warping function code. I am not sure about how to put an assertion there though... I am open to any ideas.

Not sure about overall coverage but I'm getting 92% coverage in warping_functions.py. Most non-covered code is on the warping function plot, should I have a test on plotting as well?

mzwiessele:
Ahh ok, could you merge in devel and push your changes, as codecov seems to not have picked up on those changes, see: https://codecov.io/github/SheffieldML/GPy/GPy/util/warping_functions.py?ref=6296a2127b43e480f9aaa6d685dc5915c35b7094

beckdaniel:
Done.

mzwiessele:
The coverage is still not there I am afraid : (
https://codecov.io/gh/SheffieldML/GPy/commit/1975403c0f5df5dc877a2e6bf107f35b745d0f8a

PR #307, opened 2016-02-23
New heteroscedastic GP model, which no longer needs Y_metadata

Imdrail:
Greetings

As a part of my batchelor thesis I integrated the variational heteroscedastic GP regression model from the paper "Variational Heteroscedastic Gaussian Process Regression" into your framework.

The model itself implements the base Model and works with the same input as a standard GP regression from your framework.
It supports the functions optimize and all needed functions for optimize, and is able to predict mean and variance.

I want to ask if you can provide me support to fully integrate this, already working, model into your framework.

Due to the difference in implementations I was so far unable to use your kernels for the model and thus use a square exponential kernel of the other framework.
Furthermore I might have not fully understood the layout of your framework and therefore possibly made mistakes in the structure of my model.

Felix

### DIFF
--- a/GPy/models/new_gp_heteroscedastic_regression.py
+++ b/GPy/models/new_gp_heteroscedastic_regression.py
@@ -1,0 +1,455 @@
+# -*- coding: utf-8 -*-
+"""
+Created on Sun Oct 04 20:27:30 2015
+
+@author: Felix
+"""
+from ..inference import optimization
+from ..core.parameterization import Param
+from ..core.model import Model
+import numpy as np
+import logging
+from numpy.linalg.linalg import LinAlgError
+import math
+
+
+logger = logging.getLogger("GPHeteroscedasticRegression")
+
+
+class GPHeteroscedasticRegression(Model):
+    """
+    Gaussian Process model for heteroscedastic regression
+
+    :param X: input observations
+    :param Y: observed values
+    Other parameters are there to give it the same interface as other Models but are not used
+    This model uses its own Square Exponentional kernel
+    """
+    def __init__(self, X, Y, kernel=None, varKernel = None, optimize = True):
+        """
+        Initializes all values and links the parameters which is needed for the optimizer array
+        """
+        super(GPHeteroscedasticRegression, self).__init__("GPHeteroscedasticRegression")
+        self.Y = Y
+        self.X = X
+        _, self.output_dim = self.Y.shape
+
+        self.num_data, self.input_dim = self.X.shape
+        self.kernel  =  SEKernel(self.input_dim)
+        self.varKernel  = SEKernel(self.input_dim)
+        self.mean = 1
+
+        self.n,self.D = X.shape
+        self.Lambda0 = np.ones([self.n,1])*math.log(0.5);
+        self.theta1 = np.ones([self.input_dim+1])
+        self.theta2 = np.ones([self.input_dim+1])
+
+        self.Lambda0 = Param('Lambda0', self.Lambda0)
+        self.theta1 = Param('theta1', self.theta1 )
+        self.theta2 = Param('theta2', self.theta2 )
+        self.mean = Param('mean', self.mean)
+
+        print("link hyperparameters")
+        self.link_parameter(self.Lambda0,index=0)
+        self.link_parameter(self.theta1,index=1)
+        self.link_parameter(self.theta2, index = 2)
+        self.link_parameter(self.mean, index = 3)
+
+    def train(self, X, Y):
+        """
+        Takes new X and Y to train.
+        """
+        self.X = X
+        self.Y = Y
+        if X.size == 0 or Y.size == 0:
+            return
+        self.n,self.D = X.shape
+        self.Lambda0 = np.ones([self.n,1])*math.log(0.5);
+        self.Lambda0 = Param('Lambda0', self.Lambda0)
+        self.link_parameter(self.Lambda0,index=0)
+        if self.optimize:
+            self.optimize_restarts(num_restarts=self.num_restarts, robust=True)
+
+    def optimize(self):
+        """
+        Optimizes by calling optimize_without_kernels and optimize_with_kernels with default steps in each.
+        """
+        self.optimize_without_kernels();
+        self.optimize_with_kernels();
+
+    def parameters_changed(self):
+        """
+        Method that is called upon any changes to :class:`~GPy.core.parameterization.param.Param` variables within the model.
+        In particular in the GP class this method reperforms inference, recalculating the posterior and log marginal likelihood and gradients of the model
+
+        .. warning::
+            This method is not designed to be called manually, the framework is set up to automatically call this method upon changes to parameters, if you call
+            this method yourself, there may be unexpected consequences.
+        """
+        self.setup()
+        self.derivatives(True,True)
+
+    def setup(self,verbose =0):
+        if verbose > 0:
+            print("Setup and calculate objective function")
+        self.Lambda = np.exp(self.Lambda0);
+        self.Kf = self.kernel.CovarianceMatrix(self.theta1,self.X)
+        self.Kg = self.varKernel.CovarianceMatrix(self.theta2,self.X)
+        sLambda = np.sqrt(self.Lambda);
+        self.cinvB = np.linalg.solve(np.transpose(np.linalg.cholesky(np.eye(self.n) + self.Kg*np.dot(sLambda,np.transpose(sLambda))).T),np.eye(self.n));   # O(n^3)
+        self.cinvBs = self.cinvB*(np.dot(np.ones([self.n,1]),sLambda.T));
+        self.beta = (self.Lambda-0.5)
+        mu = np.dot(self.Kg,self.beta)+self.mean
+        
+        hBLK2 = np.dot(self.cinvBs,self.Kg);
+        self.Sigma = self.Kg - np.dot(np.transpose(hBLK2),hBLK2);
+        
+        R = np.exp(mu-np.transpose(np.array([np.diag(self.Sigma)/2])));
+        p = 1e-3; # This value doesn't affect the result, it is a scaling factor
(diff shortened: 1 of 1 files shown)


lawrennd:
@alansaul, sorry, know you're busy right now but this relates a bit to your
AISTATS stuff. Is that integrated? Does this fit in with it?

@ric70x7 this is somewhat orthogonal to your heteroscedastic material, but
probably worth you also taking a look if you've time!

alansaul:
Certainly very relevant yes @lawrennd, it is one of the papers that our AISTATS paper builds off. It is the less-scalable version (O(n^3)), that isn't considering non-Gaussian likelihoods. It has a much less difficult optimization task though (great for smaller data) as it collapses out q(f), and has a nice trick which allows you to dramatically reduce the number of variational parameters by exploiting a common diagonal matrix in the posterior covariance, sadly we can't exploit this ourselves. My code isn't yet integrated, it's about to be released in another repository as soon as I get round to uploading the Arxiv paper (very soon).

Thanks for your pull request @Imdrail, we really appreciate your efforts! From a first quick look:

I don't think there is any fundamental issue that would not allow us to use the existing kernels in GPy, but some changes will need to be made essentially:

The functions CovarianceMatrix and TestSetCovariances should be the same as our kernels K function, we just take two sets of X's (X,X), or (X,X_) or (X_,X*) for example, and compute the covariance, the hyperparameters are held within the kernel object itself (so it can already see them and they don't need to be passed), and when you call:
self.link_parameter(self.kernel)
self.link_parameter(self.varKernel)
the model will be able to access these hyper parameters from its children kernels if that makes sense? using the notation, self.kernel.lengthscale and self.varKernel.lengthscale, and self._param_array_ will already contain their values. This bit should be simple.

Kernels derivatives are a little bit more complicated so this is probably where you were tripped up by our framework.
Our kernels have a function, called in the following way, self.kern.update_gradients_full(dL_dK, self.X), in parameters_changed. Given you have determined the matrix dL_dK (where L is the marginal likelihood, in your case F), we will apply the chain rule for each parameter:
dL_dtheta_i = dL_dK x dK_dtheta_i
for all of the covariance functions hyper-parameters, theta. This can be a bit of a pain to get your head around at first (especially determining what dL_dK is), let us know if this isn't clear. 

Unless I'm mistaken, the functions optimize_without_kernels, optimize_with_kernels, objective_function_gradients_without_kernels, objective_function_gradients_with_kernels, _objective_grads_without_kernels, _objective_grads_with_kernels, _grads_without_kernels, and _grads_with_kernels shouldn't be necessary. I believe these are a substitute for fixing certain parameters whilst optimizing others, something that we can do very naturally within GPy.
Parameters can be fixed using m.name_of_parameter.fix(), when this is done, the optimizer will not try and move these parameters during optimization.
For example if we had something like:
m = GPHeteroscedasticRegression(X,Y)
m.theta1.fix()
m.theta2.fix()
m.optimize()
and
m = GPHeteroscedasticRegression(X,Y)
m.Lambda0.fix()
m.mean.fix()
m.optimize()

The first will fix the kernels hyperparameters and only optimize the variational parameters. The second will fix the variational parameters and only optimize the hyper parameters. If I am understanding the code correctly, this is what the above functions are doing also but more verbosely. You can unfix as easily with m.mean.unfix() etc, or assign new constraints for example m.mean.constrain_bounded(2,3)

When integrated with the GPy kernel functions, you it would be more natural as you could do:
m = GPHeteroscedasticRegression(X,Y)
m.kernel.fix()
m.varKernel.fix()
m.optimize()
or
m.kernel.lengthscale.fix()  # to only fix the lengthscale but leave the variance. It would also open up all of the kernels available in GPy for your use.

Finally it would be ideal if you could provide some tests to ensure that everything is working correctly. We try and maintain a reasonable coverage which is beginning to slip so we want to try and keep it up.

Thanks,

Alan

Imdrail:
Hi Alan,

I will fix the points which have nothing to do with the kernels first within the next days and then fix the kernels and I will ask if i get to a point where I have problems in understanding.
Thank you very much for the answer and especially the trick with param.fix() is helping me extremely as for cleaning up code

Thanks
Felix

mzwiessele:
Hi Imdrail that looks very promising, though the checks fail. Additionally, you want to merge into master, which is not preferred for this project. Could you pull from devel (checkout the devel branch) and create a new pull request for devel?

Imdrail:
@alansaul 
While trying to implement your kernels I noticed, that the kernel i am using is having a fundamentally different parameter calculation.

Basically the results from both kernels are the same, if the lengthscale is given to e^x and the variance to e^2x.
The derivate of the parameters is also not divided through the corresponding lengthscale or variance.
I upload a file here containing my kernel and where the difference between the kernels is shown.
Only an installation of GPy is needed to run it.

For a quick workaround in the framework I tried changing the parameters of the kernels back and forth every time they are used, but unfortunately this results in a call to parameters_changed() which results in an infinite recursion

[testKernels.zip](https://github.com/SheffieldML/GPy/files/161705/testKernels.zip)

@mzwiessele 
Do I unerstand it correctly, that I should delete the actual pull request and form a new one or is there a way to change the actual pull request to the settings you require

Greetings 

Felix

mzwiessele:
I think you will benefit from creating anew pull request, started from devel, as you will otherwise have to merge into devel. 

The easiest way is to copy your files locally, then checkout devel and put your files in again. Then you delete the old pull request and create a new one from there.

mzwiessele:
bump

Imdrail:
i wanted to wait until @alansaul is answering before I close this thread since i really need these answers ._.
@mzwiessele  Should I either wait or open the new repository and leave this one or open the new one and repost my post.

Greetings
Felix

mzwiessele:
Oops sry, alan can you reopen?

alansaul:
Hi sorry @Imdrail I completely missed this.

In your code, in order to maintain the fact that the lengthscales and variances must be positive, you rightly have to exponentiate the values of theta1 (as the optimiser could give you negative values of theta, and lengthscales must be positive). In this case your lengthscale 'parameter' is whatever the exponential of theta is. 

In GPy we essentially do this for you automatically for parameters that we expect to be positive. The kernels parameters (as defined here: https://github.com/SheffieldML/GPy/blob/devel/GPy/kern/src/stationary.py#L73) are lengthscales and variance can be anything. We ensure positivity by applying an log1p(exp(lengthscale)) function just before passing it to the optimizer (note the LogExp argument in the linked code above). Whatever result the optimiser gives, we then untransform it, log(exp(optimizer_value) - 1). This comes from the parameters "constrain_positive" constraint. You could do this to see:

```
print(gpyKernel) # notice it has constraints for lengthscale and variance as +ve, meaning constrain it positive by transforming it with a exp once it comes out of the optimizer
gpyKernel.unconstrain() #Removes all constraints, so we now no longer touch it before giving it to the optimizer, and don't exp it when it comes out either
print(gpyKernel)
gpyKernel.lengthscale.constrain_positive()
gpyKernel.variance.constrain_positive()
print(gpyKernel) # back to being constrained again
# You can also use things like gpyKernel.lengthscale constrain_fixed(1.5) if you don't want to optimize the lengthscale, but just optimize the variance.
```

Essentially, I believe our implementations are the same, however ours is parameterised in terms of the lengthscale and variance, and yours is parameterised (theta1) in terms of the log of the lengthscale. As for variance, we use the same trick, you however do exp(2*variance), I'm not sure why you do this multiplication by 2.

This would explain why K is equal when you use gpyKernel.lengthscale = np.exp(theta1[:D]), gpyKernel.variance = np.exp(theta1[D]).

For you, theta1[:D] is the log of the lengthscale, for us, we expect to be given the lengthscale in its untransformed state, i.e. exp(theta1[:D]) for you. Again I'm not sure why you parameterise the variance as np.exp(2_theta1[D]), that appears to me like that you are parameterising the model by 0.5_(real_signal_variance) but I could be wrong.

I can't quite decrypt what you are doing under # derivate Functions but essentially it should be the same, gpyKernel.gradient should give back the gradients, in the space of the untransformed parameters. I'm not 100% about the missing /lengthscale and missing /variance, I would need to run through the maths a bit more thoroughly, our code for gradients of the kernel is quite well tested though (numerically), so I'm inclined to think the mismatch is todo with the parameterisation mentioned above.

The reason for the negative however, is likely because in your model you are setting the objective function, self.obj_function, which is the negative of the log_likelihood. For gaussian process models, we typically inherit from GP, rather than Model, and for these models, self.objective_function is the negative of the log_likelihood of the model (https://github.com/SheffieldML/GPy/blob/devel/GPy/core/model.py#L29), as we wish to minimise this quantity, in order to maximise the log_likelihood, third party optimisers typically minimise rather than maximise. 

Ideally you would also inherit from GP, as it is a GP model. All you would need to do is self._log_marginal_likelihood = F instead of self.obj_function = -F and similarly for the gradients (https://github.com/SheffieldML/GPy/blob/devel/GPy/core/model.py#L48), this isn't a huge issue if you prefer to continue inheriting from Model, it would just be more natural to inherit from GP (since it is one). When inheriting from a GP, the gradients of the parameters that need to be set, (gpyKernel.lengthscale.gradient) expect to be the derivative of the log_likelihood wrt to the parameter, rather than the derivative of the objective_function wrt to the parameter when inheriting from the Model if that makes sense.

Does that clarify a few things? If the gradients still make no sense after noting that we have a different parameterisation of the lengthscale and variance, let me know and I will try and take a more thorough look at the part under # derivate Functions, but I suspect that it will become more clear with the above information.

mzwiessele:
Any news?

Imdrail:
Sry the deadline for my paper is next week after that i will look into the code and hopefully fix this but at this moment I have an awful lot to do

ekalosak:
@Imdrail Can you rebase this on `SheffieldML:devel` HEAD? It looks good, the GPy CI has been updated and is working, and if we can get this to pass it LGTM to merge.

ekalosak:
Closing due to inactivity.

PR #308, opened 2016-02-24
Python3.5 fix for next()

Luk0r:
Python3.5 fix for next()
http://stackoverflow.com/questions/20484195/typeerror-range-object-does-not-support-item-assignment


mzwiessele:
Fixed on devel branch

Issue #253, opened 2015-09-16
Variances can go (numerically) negative


lionfish0:
Mainly a problem for plotting.

Variances can become negative, which causes plots to 'flip'?

(unclear if the problem goes beyond plotting, ask Alan).

mzwiessele:
Could you please provide an example script? So we can reproduce the error?

mzwiessele:
I thought this was fixed actually, as there is always constant (infinitesimal) jitter on the inverses. Please provide some example, then we can check how that happens.

alansaul:
Sorry I forgot about this until I ran into it again. This happens with numerical inaccuracies I believe, but in some cases its drastic and we should catch it (see example below).

Both noise variance and Gaussian noise are learnt to be very small, K will get numerically unstable (small negative eigenvalues in the following example). We should be able to deal with this when predicting, at least by adding jitter. In the example we below, we have (almost) identical observations, which will mean the matrix isn't full rank, or is very close to not being (minimum eigenvalue of 9.99999892865e-08).
To be honest, I'm not sure how this matrix is even being cholesky decomposed, as the jitter we add is not random, it is connected to the mean of the diagonal of K:
https://github.com/SheffieldML/GPy/blob/devel/GPy/util/linalg.py#L66
I have had similar effects occur when we don't have duplicate observations, I will try and find another example, so we need to figure out a way to resolve it for any case that K may become unstable.

Of course we could add a larger white variance, but I think we should at the very least catch these issues, and a best try and figure out how we can resolve them.

```
import matplotlib
import numpy as np
import GPy
import matplotlib.pyplot as plt
np.random.seed(0)
X = np.linspace(0,5)[:, None]
X = np.vstack((X, X[-10:,:]))
Y = np.sin(X) + np.random.randn(*X.shape)*1e-5

SE = GPy.kern.RBF(input_dim=1) + GPy.kern.White(1, variance=1e-7)
kernel = SE
m = GPy.models.GPRegression(X, Y, kernel)
m.kern.white.fix()
m.optimize()
m.plot()
mean, var = m._raw_predict(m.X)
print(var.min())
print(np.diag(m.kern.K(m.X)))
print(np.linalg.eigvals(m.kern.K(m.X)).min())
```

mzwiessele:
[...]

Issue #154, opened 2014-11-03
metropolis hastings does not work


jameshensman:

mzwiessele:
Updates?

mzwiessele:
Who wrote the metropolis hastings algorithm, and can we have this updated?

mzwiessele:
Hello?

mzwiessele:
@zhenwendai Can we close this, by either deleting hmc or implementing it?

zhenwendai:
I think we could either close this and delete the mcmc sampler code or 
implement this (if somebody really need it).

Zhenwen

lawrennd:
@zhenwendai @mzwiessele ... don't delete hmc ... that works right? This is
for metropolis hastings right?

zhenwendai:
hmc sampler works fine and we have test cases for that. This is for the 
old metropolis hasting implementation that hasn't adapted to the param 
framework.

Zhenwen

lawrennd:
I'd agree on removal then as a tidying up step.

mzwiessele:
Is it deleted? Can we close the issue? @zhenwendai

zhenwendai:
Not yet. I can do the deletion in the fix negative prediction pull 
request that I will get it done tomorrow.

Zhenwen

mzwiessele:
Is the pull request in? It still seems to fail...

PR #314, opened 2016-02-29
Fix the issue of negative predicted variance of normal GP

zhenwendai:
1. Fix the issue of negative predicted variance of normal GP
2. get rid of mpi4py when import
3. fix gpu initialization


mzwiessele:
@zhenwendai could you add tests to run the code, so that the coverage is provided for these changes?

zhenwendai:
Sure. One of the changes is only an interface. There is no real runnable 
code in GPy. I will do for the rest.

Zhenwen

mzwiessele:
Do we have an update on this?

mzwiessele:
I recently tried to run the SSGPLVM example and it seemed to fail sporadically. Can we get some tests, testing the gradients and run the code inside the tests, so that we can have this good work for the future and versions to come?

zhenwendai:
I can add test cases for gradient checks. One problem of spike and slab 
model is that it depends a lot on initialization. With random 
initialization, it might fail randomly.

Zhenwen

mzwiessele:
Great, gradient checks are fine to work with, at least the code gets run then. Try to check branching code, like if statements thoroughly, so that we dont get surprised by changes. You might want to have a look at the mnibatch_tests.py. 

There is a great package called coverage to check coverage locally. When installed you can run from the root directory:

coverage run travis_tests.py
coverage html

and it will create a coverage report for you. You can see the coverage of all files then in htmlcov directory. This also shows the branching coverage and unhandled special cases.

mzwiessele:
Updates? Is this the new pull request, or still the old one?

zhenwendai:
@mzwiessele 
Sorry for the slowness. 
1.  I have added a testcase for SSGPLVM.
2.  The metropolis hastings code is new and compatible with the param framework (I didn't know before). I did a bug fix. It can run now, but I didn't check the correctness of the results. I added a testcase that makes sure it is runnable.

fixes #253
fixes #154

mzwiessele:
Thanks so much! There is copied code now between sparse gp and normal gp predict functions. is that necessary? Can we not get those together into one, so that when we make changes they are reflected in both sparse gp and gp?

zhenwendai:
It is different from the default sparse gp prediction, because the 
solution of the numerical problem is to solve the covariance matrix in 
the fly. Unfortunately, the prediction code can only be shared between 
normal gp, sparse gp and Bayesian GPLVM if we store the inversion of 
woodburry matrix (the numerical instable way). Therefore, I implement 
the _raw_predict function in the posterior object, so that different 
model can have different numerical stable implementation.

Zhenwen

mzwiessele:
Ahh nice, I like it! So we can get rid of the prediction code in GP and SparseGP and put it in the super class of Posterior?

zhenwendai:
Cool! I have done the modification correspondingly. I have moved the _raw_predict function into posterior object. 

The modification of the testcase is to allow the behavior of clipping the negative predicted variance.

PR #324, opened 2016-03-08
[kern] Standard periodic kernel. Changes parameter name from 'waveleng…

AlexGrig:
[kern] Standard periodic kernel. Changes parameter name from 'wavelength' to 'period'.

This seems to be more clear. Also some minor modifications in the same file.

### DIFF
--- a/GPy/kern/src/standard_periodic.py
+++ b/GPy/kern/src/standard_periodic.py
@@ -1,6 +1,5 @@
 # -*- coding: utf-8 -*-
-
-# Copyright (c) 2014, GPy authors (see AUTHORS.txt).
+# Copyright (c) 2015, Alex Grigorevskiy
 # Licensed under the BSD 3-clause license (see LICENSE.txt)
 """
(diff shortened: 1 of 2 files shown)


mzwiessele:
I like the lengthscale, but why is period better then wavelength?

AlexGrig:
Hi, Max

I think that when you deal with periodicities, it is more standard to
think in terms of frequencies and periods.
Wavelength is more related to waves - periodicities is both time and
space. Hence, I think for average
person it is easier to deal with period. For instance if you need to
convert it to frequency it is just 1/(period),
but for wavelength you need to remember or derive a formula.

Lengthscale is already used in this kernel for other parameter, so I
can't use this again

Br,
    Alex.

mzwiessele:
I meant the renaming from lengthscales to lengthscale, good catch there : )

Ok, I’ll take it, period it is then : )

I always thought of wavelength and period to be synonyms, as I do not know much in that field. 

Thanks

PR #345, opened 2016-03-20
included X gradient for standard periodic

sachinruk:
I think this will do for the standard periodic. Will include more changes if this is correct.

I did add some commented code in the poly kernel as well, but I'm not 100% sure about that. Especially if X is none what I should do to the diagonal elements of dK_dX. If dK_dX is a NxD matrix then I'm asking the wrong question as well.

### DIFF
--- a/GPy/kern/src/poly.py
+++ b/GPy/kern/src/poly.py
@@ -57,3 +57,30 @@ class Poly(Kern):
 
     def gradients_X_diag(self, dL_dKdiag, X):
         raise NotImplementedError
+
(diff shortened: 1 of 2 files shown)


sachinruk:
err was I supposed to do something apart from editing the code? Not sure what this travis-ci thing does.

mzwiessele:
the travis ci checks the tests. thus your gradients work, as our kernel gradient suite (if im not completely mistaken) checks all gradients that are active.

sachinruk:
I think I'll run the nosetests before I contact you again. Apologies for taking up your time.

PR #350, opened 2016-03-24
Fixed inputs and BGPLVM prediction tests

alansaul:
Added a convenience function for calculating an array suitable for passing to "fixed_inputs" when making predictions. Fixed inputs can be chosen to be computed to be the mean, median, or zeros.

Also Andreas mentioned he was having trouble with the BGPLVM prediction with uncertain inputs. Although this appears to be working correctly, we didn't have a test for the moment matching. Added a test for a linear kernel, in this case the output is still Gaussian, so the moment matching should discover the correct moments, and predictions should be close to those being projected.


mzwiessele:
Good work! Could you get the added code into coverage as well, so we can merge it in?

Issue #352, opened 2016-03-30
Laplace inference not working with binomial likelihood


djsutherland:
Laplace inference appears to be broken with binomial likelihoods. [Here](https://gist.github.com/e34d5e3e31e3d7a08a6e6109fe5a3ec5) is a notebook demonstrating the problem.

Specifically, with 1000 inputs, it's trying to allocate a 1000000 x 1000000 matrix, which of course fails since my machine doesn't have 7 terabytes of free memory. But that matrix should only be 1000 x 1000. Looking into it a little bit, it seems that the `W` matrix made [here](https://github.com/SheffieldML/GPy/blob/6ca841f35be5dc3f03aff20b250127047b70351f/GPy/inference/latent_function_inference/laplace.py#L184), which should be 1000 x 1, is actually 1000 x 1000, with each of the rows apparently identical.

This doesn't happen with a Poisson likelihood, so the error is probably in `BinomialLikelihood.d2logpdf_df2`. But I haven't tried to dig into those functions further.

mzwiessele:
@alansaul is that right? Maybe there is the problem of what you where facing?

alansaul:
Oops this must have slipped through the testing net. I believe this particular memory explosion issue might actually be with the shape of the Y_metadata, it must be the same shape a Y:

```
Y_metadata={'trials': ns[:, None]}
```

We should obviously do a check for this, I will add a check and a test case. However it seems like we have never made use of Laplace with the Binomial likelihood since d3logpdf_dlink3 has never been implemented. If you would like to make a pull request that would be great, if not, I don't mind doing it when I get a minute next week.

Sorry for the inconvenience! Alan

alansaul:
Scratch that, I found a minute. @dougalsutherland I haven't tested that this gives sensible results, can you take a look with an example you expect to work and see if it does anything reasonable? Sampling may not work yet so to generate predictions for Y you may need to change "samples" in likelihoods/binomials.py

djsutherland:
Thanks! Personally, I find requiring 2d `Y` extremely counter-intuitive, and didn't even consider that.

Comments about #356 over there.

mzwiessele:
Well, that is why we make you do it! It is mathematical the right thing to give a column vector into a mathematical framework. You should get used to these things, as as soon as you start working with mathematical code, which allows for dimensions it will decrease your rate of error significantly!

djsutherland:
[...]

PR #356, opened 2016-04-01
Binomial laplace #352

alansaul:
Added derivative for Binomial, corrected some maths, and added test. As requested by #352


djsutherland:
Thanks! I tried actually running my stupid example from the previous notebook [here](https://gist.github.com/9ab8a75177c47f12e6625c5a1f2661b3):
- Basic modeling seemed fine.
- When I tried to optimize the model, the lengthscale went really small and it blew up, but this is probably a Laplace issue and not a code issue....
- `_raw_predict` works.
-  `predict` doen't: it crashes inside the quadrature in `predictive_mean`, because [this line](https://github.com/SheffieldML/GPy/blob/378ac0ad06919d10f2805b8546737596722ea3ac/GPy/likelihoods/likelihood.py#L303) calls `self.conditional_mean(f)` without passing `Y_metadata`. It seems like probably every likelihood needs to be changed to take an optional `Y_metadata` argument, and this line (and presumably similar ones) changed to pass those through.

mzwiessele:
Could you merge in the current devel branch, to see if that is the missing coverage here? Otherwise it looks very good : )

djsutherland:
So, I made the changes I suggested above to the API of `conditional_mean` (in https://github.com/dougalsutherland/GPy/commit/7c3ade9fc5bcf5d22bdfaa8990ed52cb103ffd88), but then noticed that https://github.com/SheffieldML/GPy/commit/ff9ab62db2057a54c2634d308eef21c82e13ffd8 (not included in your branch here, since it was merged to devel just the other day) removed the `BinomialLikelihood.conditional_mean` function entirely, which seems to make `predict` use sampling rather than quadrature. Is there a reason for this, or another way I should be making predictions with a model like this? (@jameshensman?)

mzwiessele:
Great stuff!! when we have fixed the coverage we can start merging 👍

mzwiessele:
Any news on this?

djsutherland:
Bump. I'd really like to distribute some analysis that uses binomial likelihoods, but would also much prefer to just have a requirement on GPy than tell people they need to check out a branch several versions behind the current one....

PR #362, opened 2016-04-06
FIX: Fixing bugs and innacuracies in state-space inference.

AlexGrig:
Correcting some inaccuracies which emerged during testing.


mzwiessele:
I had a look and you are indeed right with your Exponential and we will have to change the lengthscale in GPy instead of in your code : ) Thanks!

mzwiessele:
I am fixing the stuff on the state_space branch

mzwiessele:
Fixing in #364

PR #374, opened 2016-04-20
Changes in kern.gradients_XX

alessandratosi:


mzwiessele:
- When I compare the old code (cov=False), it is not the same : (
- comparing the diagonal of cov=True, to cov=False it is not the same either
- the wishart_embedding has nans in it, and is far of what it was before

**Old**:
figure_1

**New**:
figure_1_new

alessandratosi:
In the previous implementation of the wishart_embedding the variance of the predict_Jacobin was wrong because the cross-covariance terms between the input dimensions Q were not computed. This has been fixed with the last commit.
I think that the error in the previous plot is coming from a bug in gradients_XX(...,cov=True), which is assigned to be zero somewhere (probably in the slice_operations?) even if the function is computing the correct values. Any chance you could have a look at that, @mzwiessele ?

mzwiessele:
But even if cov is false, the resulting wishart embedding is different from the current one on devel! I am having a look at the setting zero issue, but we need to find the problem with the above.

alessandratosi:
Sure, I'll look into that.

mzwiessele:
Wait, I have fixed the slicing and now the diagonals are the same:

``` python
t = m.kern.gradients_XX(np.ones((1,1)), m.X.mean, cov=True)
f = m.kern.gradients_XX(np.ones((1,1)), m.X.mean, cov=False)

(t.T[np.diag_indices(m.input_dim)].T - f).sum()
>>> 0.0
```

GPy/core/gp.py
```diff
@@ -378,27 +373,28 @@ def predict_jacobian(self, Xnew, kern=None, full_cov=True):
             dK_dXnew_full[i] = kern.gradients_X(one, Xnew, self._predictive_variable[[i]])
 
         if full_cov:
-            dK2_dXdX = kern.gradients_XX(one, Xnew, cov=False)
+            dK2_dXdX = kern.gradients_XX(one, Xnew)
         else:
```
mzwiessele, line 379: Full cov is the covariance across samples here! It has nothing to do with dimensions...

GPy/kern/src/stationary.py
```diff
@@ -218,58 +218,74 @@ def gradients_X(self, dL_dK, X, X2=None):
...
-        #grad = np.empty(X.shape, dtype=np.float64)
-        for q in range(self.input_dim):
-            tmpdist2 = (X[:,[q]]-X2[:,[q]].T) ** 2
-            grad[:, :, q] = ((tmp1*invdist2 - tmp2)*tmpdist2/l2[q] - tmp1)/l2[q]
-            #grad[:, :, q] = ((tmp1*(((tmpdist2)*invdist2/l2[q])-1)) - (tmp2*(tmpdist2))/l2[q])/l2[q]
-            #np.sum(((tmp1*(((tmpdist2)*invdist2/l2[q])-1)) - (tmp2*(tmpdist2))/l2[q])/l2[q], axis=1, out=grad[:,q])
-            #np.sum( - (tmp2*(tmpdist**2)), axis=1, out=grad[:,q])
+            #tmp1[X==X2.T] -= self.variance # Old version, to be removed 
+                                            # (seems to have a bug: it is subtracted to the first X1 anyway)
+            tmp1[invdist2==0.] -= self.variance
```
mzwiessele, line 255: What does that even mean? invdist is the INVERSE distance, so if anything you need to check this for infinite entries...

GPy/kern/src/stationary.py
```diff
@@ -218,58 +218,74 @@ def gradients_X(self, dL_dK, X, X2=None):
...
+        
+        if cov: # full covariance
+            grad = np.empty((X.shape[0], X2.shape[0], X2.shape[1], X.shape[1]), dtype=np.float64)
+            for q in range(self.input_dim):
+                for r in range(self.input_dim):
+                    tmpdist2 = (X[:,[q]]-X2[:,[q]].T)*(X[:,[r]]-X2[:,[r]].T) # Introduce temporary distance
```
mzwiessele: We can compute X-X2 in q before the r loop, so we do not have to compute it for every other dimension over and over again

GPy/kern/src/stationary.py
```diff
@@ -218,58 +218,74 @@ def gradients_X(self, dL_dK, X, X2=None):
...
+                    if r==q:
+                        grad[:, :, q, r] = np.multiply(dL_dK,(np.multiply((tmp1*invdist2 - tmp2),tmpdist2)/l2[r] - tmp1)/l2[q])
+                    else:
+                        grad[:, :, q, r] = np.multiply(dL_dK,(np.multiply((tmp1*invdist2 - tmp2),tmpdist2)/l2[r])/l2[q])
```
mzwiessele: Also there has to be a better way of writing this, looping through q**2 seems to be too much, what do you think?

GPy/core/gp.py
mzwiessele, line 379: This is where the bug starts with the wishart embedding, btw.

GPy/kern/src/stationary.py
```diff
@@ -218,58 +218,74 @@ def gradients_X(self, dL_dK, X, X2=None):
...
+                        Thus, we return the second derivative in X2.
         """
-        # The off diagonals in Q are always zero, this should also be true for the Linear kernel...
         # According to multivariable chain rule, we can chain the second derivative through r:
         # d2K_dXdX2 = dK_dr*d2r_dXdX2 + d2K_drdr * dr_dX * dr_dX2:
         invdist = self._inv_dist(X, X2)
         invdist2 = invdist**2
-
-        dL_dr = self.dK_dr_via_X(X, X2) * dL_dK
+        dL_dr = self.dK_dr_via_X(X, X2) # * dL_dK we perofrm this product later
```
mzwiessele: Why not do the product here?

GPy/kern/src/stationary.py
alessandratosi, line 255: tmp1[X==X2.T] is subtracting self.variance from the first vector of X anyway, so it has a bug (in my understanding) and I fixed it. 
I have replaced this with tmp1[invdist2==0.] -= self.variance ignored to subtract the term self.variance anytime a vector of X is equal to a a vector of X2. Also Remember that if the distance is zero then the function _inv_dist is returning ZERO as well 

```
def _inv_dist(self, X, X2=None):
    """
    Compute the elementwise inverse of the distance matrix, expecpt on the
    diagonal, where we return zero (the distance on the diagonal is zero).
    This term appears in derviatives.
    """
    dist = self._scaled_dist(X, X2).copy()
    return 1./np.where(dist != 0., dist, np.inf)
```

mzwiessele, line 255: `tmp1[X==X2.T] -= self.variance` subtracts self.variance from entries, where X and X2 are the same for all pairs of (x,x_2) in {X x X_2}

I am not sure what you mean by "is subtracting self.variance from the first vector of X anyway"

GPy/core/gp.py
alessandratosi, line 379: It has to do with dimension in this sense that here a sample is a vector of the Jacobian (one vector for each dimension Q). 

The bug in the previous version of the code was due to the fact explained in the (deleted) note:

Note: We always return sum in input_dim gradients, as the off-diagonals
-        in the input_dim are not needed for further calculations.
-        This is a compromise for increase in speed. Mathematically the jacobian would
-        have another dimension in Q.

That _missing_ dimension Q was producing a covariance function which was zero in the off-diagonal terms relative to the input dimension, which was wrong. This error was propagated in the computation of the Wishart Embedding.

GPy/kern/src/stationary.py
alessandratosi: I checked again and the old code is doing the same as the new version, so I will change it back to
tmp1[X==X2.T] -= self.variance, which makes more sense, my mistake.

GPy/core/gp.py
alessandratosi, line 379: Could you push the changes i the slicing, please? So I'll check the plots. Many thanks @mzwiessele !

lawrennd:
@mzwiessele are these changes folded in?

Issue #319, opened 2016-03-04
Cannot import GPy on mybinder with Python 3.5


glouppe:
Hi,

I am trying to make public on mybinder.org some notebooks of mine where GPy is used. Installation works fine, but importing `GPy` fails with the following error:

```
/home/main/anaconda2/envs/binder/lib/python3.5/site-packages/GPy/core/__init__.py in <module>()                                    
      6 from . import parameterization                                                                                             
      7                                                                                                                            
----> 8 from .gp import GP                                                                                                         
      9 from .svgp import SVGP                                                                                                     
     10 from .sparse_gp import SparseGP                                                                                            

/home/main/anaconda2/envs/binder/lib/python3.5/site-packages/GPy/core/gp.py in <module>()                                          
      3                                                                                                                            
      4 import numpy as np                                                                                                         
----> 5 from .. import kern                                                                                                        
      6 from GPy.core.model import Model                                                                                           
      7 from paramz import ObsAr                                                                                                   

/home/main/anaconda2/envs/binder/lib/python3.5/site-packages/GPy/kern/__init__.py in <module>()                                    
      6    :private-members:                                                                                                       
      7 """                                                                                                                        
----> 8 from .src.kern import Kern                                                                                                 
      9 from .src.add import Add                                                                                                   
     10 from .src.prod import Prod                                                                                                 

/home/main/anaconda2/envs/binder/lib/python3.5/site-packages/GPy/kern/src/__init__.py in <module>()                                
----> 1 from . import psi_comp                                                                                                     

/home/main/anaconda2/envs/binder/lib/python3.5/site-packages/GPy/kern/src/psi_comp/__init__.py in <module>()                       
     66                                                                                                                            
     67 from . import ssrbf_psi_gpucomp                                                                                            
---> 68 from .rbf_psi_gpucomp import PSICOMP_RBF_GPU                                                                               

/home/main/anaconda2/envs/binder/lib/python3.5/site-packages/GPy/kern/src/psi_comp/rbf_psi_gpucomp.py in <module>()                
      6 from paramz.caching import Cache_this                                                                                      
      7 from . import PSICOMP_RBF                                                                                                  
----> 8 from ....util import gpu_init                                                                                              
      9                                                                                                                            
     10 gpu_code = """                                                                                                             

/home/main/anaconda2/envs/binder/lib/python3.5/site-packages/GPy/util/__init__.py in <module>()                                    
      7 from . import squashers                                                                                                    
[...]

PR #378, opened 2016-04-22
suppress UnicodeDecodeError: ascii codec - when import GPy

ken-okabe:
#319 : Cannot import GPy on mybinder with Python 3.5

The change is to give 'utf-8' encoding option of the function in `detasets.py`
My environment is on console and JupyterNotebook with Python 3.5.


kayhan-batmanghelich:
@kenokabe I installed the latest version where your modification is already applied. I still have issue (#387). Any idea how to resolve this?

PR #431, opened 2016-08-23
LogisticBasisFuncKernel fix

bwengals:
Removed Logsumexp() from LogisticBasisFuncKernel, allowing slope parameter to be negative.  Also removed unnecessary scipy import.


mzwiessele:
Great stuff, btw. for now you can just manually remove the constraint after creating the model:

```
m.kern.slope.unconstrain()
```

bwengals:
Ah, awesome didn't know that.  Thanks for the tip!

PR #448, opened 2016-10-19
Added pep.py -- Sparse Gaussian processes using Power Expectation Propagation

thangbui:
This approach allows interpolation between FITC (EP or alpha = 1), and Titsias's variational (VarDTC, VFE when alpha = 0).

Reference: A Unifying Framework for Sparse Gaussian Process Approximation using Power Expectation Propagation https://arxiv.org/abs/1605.07066


lawrennd:
Wow! That's super cool Thang!

@mzwiessle, @zhenwendai it would be great to get this incorporated as soon
as feasilble. @zhenwendai do you have time to take a look?

Neil

thangbui:
Hi All,

I also provide a script comparing different techniques here https://github.com/thangbui/sparseGP_powerEP/blob/master/python/sgp/tests/test_PEP_GPy.py

Thanks,
Thang.

mzwiessele:
Great stuff! Could you add the test script as a test in GPy so the code is being tested within the GPy tests?

thangbui:
@mzwiessele I added a test script which checks grads (as in fitc) and makes sure special cases of PEP are FITC and VarDTC -- not sure if this is enough.

zhenwendai:
Great work! Thanks a lot, @thangbui! The test case of PEPgradienttest sometime fails randomly because of the random generated dataset. Could you fix the seed? 

The codecov does not get updated with the recent commit. I don't know why.

zhenwendai:
@mzwiessele, could you have a look at why the added test case is added but the coverage does not increase? I am not familiar with codecov.

mzwiessele:
Sure as soon as I have recovered, Ill give it a thorogh look.

mzwiessele:
It is simple: Tests for nosetests need to have "test" in their name. pep.py is not being run as a test....

thangbui:
Thanks @mzwiessele and @zhenwendai for the merge!

mzwiessele:
Thanks for the contribution! Looking forward to the next module : )

Issue #404, opened 2016-06-23
Plotly offline plotting option


mikeschaekermann:
I would love to have an option for using Plotly in offline mode so that I don't have to create a user account on Plotly and graphs do not automatically get uploaded to the Plotly cloud. On the one hand, I love Plotly and all its beautifully designed plotting functions, but on the other hand I cannot use it for some of the sensitive data that I am working with because this data may not be transferred to external servers.

Offline plotting in Plotly: https://plot.ly/python/getting-started/#initialization-for-offline-plotting

mzwiessele:
We are very happy to see people engaged in this. If you want to give it 
a shot, put up a pull request with the option, see

https://help.github.com/articles/fork-a-repo/
and
https://help.github.com/articles/creating-a-pull-request/

mzwiessele:
Any progress updates?

mikeschaekermann:
Sorry, I haven't found the time to do this yet. Currently quite busy with a work project.

alexfeld:
I've just added a pull request that I believe addresses this issue. The offline plots mostly work for me albeit with a few issues. I think these issues are `plotly.py`'s though.

PR #480, opened 2017-02-22
add offline plotting for plotly

alexfeld:
To address #404

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,6 @@ iterate.dat
 # git merge files #
 ###################
 *.orig
+
+# pycharm IDE stuff
+.idea/
--- a/GPy/plotting/__init__.py
+++ b/GPy/plotting/__init__.py
@@ -2,7 +2,7 @@
 # Licensed under the BSD 3-clause license (see LICENSE.txt)
 current_lib = [None]
 
-supported_libraries = ['matplotlib', 'plotly', 'none']
+supported_libraries = ['matplotlib', 'plotly', 'plotly_online', 'plotly_offline', 'none']
 error_suggestion = "Please make sure you specify your plotting library in your configuration file (<User>/.config/GPy/user.cfg).\n\n[plotting]\nlibrary = <library>\n\nCurrently supported libraries: {}".format(", ".join(supported_libraries))
 
 def change_plotting_library(lib, **kwargs):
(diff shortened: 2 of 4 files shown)


alexfeld:
I should note that there are other ways to implement this such as importing `offline` or `plotly` at the instantiation of the class itself, but I think this is the most clear way with how you've constructed the `plotting` module thus far.

mzwiessele:
Looks good, did you test it? Unfortunately I don't have access to offline plotting with plotly.

alexfeld:
I tested it using your GP regression example script and it worked as intended with the `plotly.offline.plot` bugs that I mentioned. I've made a request to them to fix the errors. It looks like they can be fixed by only passing `numpy` arrays of the shape `(n,)`, but hopefully they fix this on their end.

mzwiessele:
We should try and make it a backend option, what do you think? We could add a third option "plotly_offline", so we do not need a global parameter. This would then call the library with the offline option.

alexfeld:
I've pushed a commit to address this. There are now three classes in the `plotly_dep.plot_definitions` module. There is a `PlotlyPlotsBase`, `PlotlyPlotsOnline`, and `PlotlyPlotsOffline`. The base class is identical to the old `PlotlyPlots` class but with the `show_canvas` function set to require implementation by a child class. This method is the only thing implemented by the `Offline` and `Online` classes and calls the respective `plot` and `iplot` functions from the `plotly` package.

Allowed inputs to the `GPy.plotting.change_plotting_library` function are `plotly` (preserves old functionality), `plotly_online`, and `plotly_offline`. I tested with the regression example and it worked as expected.

mzwiessele:
That looks awsome! I am on the problems with the new numpy (that's why the builds fail) and will merge them in here before pulling. Great work!

alexfeld:
No problem. Thanks for the great package!

mzwiessele:
Oh wait I just tried it and it actually works! I didn't know you could use plotly offline... Great stuff, holy moly.

mzwiessele:
I added a little workaround with squeeze into the plotting defs over at #482.
As soon as the PR checks it will be pulled and I will update GPy on pip.

Thanks again!

alexfeld:
Ok, sounds good! The offline plotting option is nice but as I mentioned it has some bugs. These are 100% plotly's problem though since if you export the offline plots to plotly's servers they work fine. Hopefully your squeeze fix helps.

PR #491, opened 2017-03-17
fix for parallel optimization

alexfeld:
fix `GP.optimize` method to match output from base class method

### DIFF
--- a/GPy/core/sparse_gp_mpi.py
+++ b/GPy/core/sparse_gp_mpi.py
@@ -88,9 +88,9 @@ class SparseGP_MPI(SparseGP):
     def optimize(self, optimizer=None, start=None, **kwargs):
         self._IN_OPTIMIZATION_ = True
         if self.mpi_comm==None:
-            super(SparseGP_MPI, self).optimize(optimizer,start,**kwargs)
+            ret = super(SparseGP_MPI, self).optimize(optimizer,start,**kwargs)
         elif self.mpi_comm.rank==0:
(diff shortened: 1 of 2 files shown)


GPy/core/sparse_gp_mpi.py
```diff
@@ -88,9 +88,9 @@ def optimizer_array(self, p):
     def optimize(self, optimizer=None, start=None, **kwargs):
         self._IN_OPTIMIZATION_ = True
         if self.mpi_comm==None:
-            super(SparseGP_MPI, self).optimize(optimizer,start,**kwargs)
+            ret = super(SparseGP_MPI, self).optimize(optimizer,start,**kwargs)
```
alexfeld, line 91: I'm not sure if this is the right course here. I've never used an MPI implementation so I'm unsure about how this would work. These modifications enable this class to be used with the python multiprocessing pool in the same way as the base GP class.

mzwiessele: @zhenwendai could you have a look?

mzwiessele:
Did you put some tests in place to test the parallel_optimize=True option? Just to make sure it runs in future versions and is being tested on the CI.

alexfeld:
Sorry for the delay. Yes, I did. I just added it to the PR

GPy/core/sparse_gp_mpi.py
zhenwendai: @mzwiessele, sorry that I didn't see this. It looks OK to me. It would not change the behavior of MPI.

PR #492, opened 2017-03-23
Changes in EP/EPDTC to fix numerical issues

pgmoren:
Changes in EP/EPDTC to fix numerical issues and increase the flexibility of the inference

    Changes to avoid numerical issues and improve the performance:
    - Keep value of the EP parameters between calls
    - Enforce positivity of tau_tilde
    - Stable computation of the EP moments for the Bernoulli likelihood
    - Compute marginal in the GP model without directly inverting tau_tilde

    Changes to improve the flexibility:
    - Add parameter for maximum number of iterations
    - Distinguish between alternated/nested mode
    - Distinguish between sequential/parallel updates in EP

### DIFF
--- a/GPy/inference/latent_function_inference/expectation_propagation.py
+++ b/GPy/inference/latent_function_inference/expectation_propagation.py
@@ -1,15 +1,16 @@
 # Copyright (c) 2012-2014, GPy authors (see AUTHORS.txt).
 # Licensed under the BSD 3-clause license (see LICENSE.txt)
 import numpy as np
-from ...util.linalg import jitchol, DSYR, dtrtrs, dtrtri
+from ...util.linalg import jitchol, DSYR, dtrtrs, dtrtri, pdinv, dpotrs, tdot, symmetrify
 from paramz import ObsAr
 from . import ExactGaussianInference, VarDTC
 from ...util import diag
+from .posterior import PosteriorEP as Posterior
 
 log_2_pi = np.log(2*np.pi)
 
 class EPBase(object):
-    def __init__(self, epsilon=1e-6, eta=1., delta=1., always_reset=False):
+    def __init__(self, epsilon=1e-6, eta=1., delta=1., always_reset=False, max_iters=np.inf, ep_mode="alternated", parallel_updates=False):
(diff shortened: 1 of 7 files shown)


alansaul:
Cool! Great work, I've been waiting for someone / meaning to take a look at the EP for a while! Did you notice much of a speed up in terms of inference? I noticed a while back that the existing implementation was a little on the slow side compared with other implementations I've seen. I guess any speed up would have come from the reimplementation of logPdfNormal logCdfNormal and cdfNormal, which from memory was one of the bottlenecks, as well as avoiding duplicate computations in the moment matching.

Is there any reason why you choose in your tests some specific values of the above to test against (pySols) rather than just comparing with the scipy implementation at a number of points?

Thanks!

pgmoren:
Thanks Alan! I have not done optimizations in terms of speed. There is some
speed up due to the fact that now the parameters of the EP approximation
are kept between iterations (before there was a variable to store the old
parameters but for some reason it was not used). Also, there is a flag to
do the EP updates in parallel which is less accurate but much faster when
you have a large number of data. The reimplementation of logPdfNormal
logCdfNormal and cdfNormal was done just to avoid numerical issues, but I
have not evaluated whether it leads to a significant improvement in terms
of speed. The code for these functions is based on
https://github.com/mseeger/apbsint, developed by Matthias Seeger. Regarding
the test, the reason why I chose some specific values instead of just
comparing with the scipy implementation is because I had this unit test
coded originally in scala so I directly translated it. If there is an
interest I can redo it.

Best,

Pablo.

2017-03-23 10:58 GMT+00:00 Alan Saul <notifications@github.com>:

ric70x7:
Great job Pablo!
Cheers,

Ricardo

zhenwendai:
Shall we merge the pull request?

mzwiessele:
Sure if you guys are confident, I haven't looked into it in detail. It does look good from my side, as far as I can tell!

Great work Pablo, thank you very much!

zhenwendai:
We did some benchmarking on classification. These changes should be fine. Let's merge it in.

PR #500, opened 2017-04-19
Ccd

adhaka:


GPy/testing/ccd_test_1.py
```diff
@@ -0,0 +1,132 @@
...
+        if model is None:
+            model = self.m
+        ls = self.find_likes(self.m, self.stepsize)
+        numsum = ls*(self.stepsize**self.ndims)
+        model.optimize()
+        hes = model.numerical_parameter_hessian()
+        # Laplace approximation to find marginal likelihoods of submodels, integrate out hyperparams of kernel matrix and likelihood params
+        hessum = np.exp(model.log_likelihood()) * 1 / np.sqrt(np.linalg.det(1 / (2 * np.pi) * hes))
+        print hessum, hes
+        print numsum
```
alansaul: This print statements are causing python 3 compatibility issues. I think ideally we avoid printing during tests though if possible perhaps it is better to remove them.

alansaul:
Looks good. Is there a reason why we shouldn't keep the simple test case in as well as the GP one though? I think we need to remove the %matplotlib inline line from ccd_tests.py I believe this is our bug though.

adhaka:
Thanks for the comments Alan, its a bit cleaner now, also I remember you had made some changes to the log marginal objective function, which you can merge in here.

lionfish0:
@adhaka Hi - bit late here, so hope this still makes sense.
Alan and I had a go at the CCD stuff - I've tried to summarise our confusion in a jupyter notebook.
https://github.com/lionfish0/ccd/blob/master/Test%20CCD%20complete.ipynb

(by the way, ignore the GPy model example further down - I'm just considering the simple example to start with).

lawrennd:
@alansaul @lionfish0 any reason this is still hanging?

lionfish0:
I wasn't confident I understood what the correct answers were - so I wasn't sure if I was writing correct test code...I've had a chat with Aki in the meantime and so I think I could probably finish it.
I actually might want this functionality soon, so I'll look at getting it sorted when I get back (November).

ekalosak:
Closing due to inactivity, incomplete PR description & incomplete PR name.

PR #513, opened 2017-06-12
adding some test cases for EP.. more work needs to be done, these are…

adhaka:
… some high level test cases ..


zhenwendai:
Is the pods dependency for test cases necessary? This will expand the dependent libraries for running tests.

mzwiessele:
Many datasets can be accessed through the GPy datasets, too. In case that helps.

mzwiessele:
You should merge in the current devel branch, which has some fixes in for the travis and appveyor builds.

adhaka:
Hii Zhenwen, I have removed the pods line as you suggested which was not really needed, also this request contains an important fix to the refactored EP-code. Let me know, if there are any more things which can be fixed. Also, I dont know why can't the pull request pass the checks. I feel I should know more about it. Thanks again for looking into it.

lawrennd:
We should have a discusison about GPy.datasets and pods etc in one of the
milestone meetings.

Issue #453, opened 2016-11-04
Student-t processes


javdrher:
Hello,

Last month I implemented Student-t processes in GPy (https://arxiv.org/pdf/1402.4306v2.pdf). I encountered the student-t likelihood and there seem to be some differences with what is currently available in GPy. One of them being that Student-t processes offer direct inference, and only have a slightly different predictive variance. 

The implementation was a bit too shaky to directly make a PR: I replaced the student-t likelihood locally, implemented a direct inference method for student-t and created a subclass of GP in GPy.core. I also included optimization of the degrees of freedom. If interested I can refactor the code and contribute the model. However, I would require some input on how to make it co-exist with the current student-t likelihood.

mzwiessele:
Thanks @javdrher that sounds great! What is the advantage over the existing inference? It seems to me, that is will be more efficient, as it is a specific implementation of the matter? 

@alansaul can you help with this?

javdrher:
Student-t processes are derived by assuming an inverse wishart prior for the covariance matrix, and marginalizing this prior out. This results in a likelihood that corresponds to a multivariate t-distribution. 
I should mention that the noise is no longer considered to be an independent distribution because the sum with p(f|x) makes derivation of the marginal likelihood intractable again. Instead, a white noise kernel is applied to handle noise. Doing so, the likelihood is a single MVT, so direct inference becomes feasible. I think this could be one of the differences with the current implementation, but I'm not entirely sure. 

Appendix A, Lemma 3 of the paper I linked also contains the derivation of the predictive distribution, which is also an MVT as opposed to the current implementation where its assumed to be Gaussian (I think?). The degrees of freedom of this predictive distribution do not vary on x, and it has the additional term in front of the variance which doesn't seem present in the current implementation.

sivark:
@javdrher Hello! Have you shared your implementation of Student-T processes on Github?

mzwiessele:
We are looking forward to a PR of this matter : )

javdrher:
My apologies for the delay in replying @sivark , the code is still somewhere on a backup disc somewhere, from what I remember it wasn't in a state I could make a PR right away. I'm a little busy right now, but I'll try to check it out this week. If I remember everything correctly I should be able to refactor it this week and create a PR.

lawrennd:
PR would be great. @mzwiessele these processes are particularly interesting
for small n because at that point the process scale (which is treated with
an inverse gamma) is not yet determined well. After about 10 data points I
[...]

PR #525, opened 2017-07-14
Student-t processes

javdrher:
Following #453, an implementation of student-t processes in GPy. I ended up re-writing my code entirely. Through `checkgrad()`, I noticed some problems with the gradient for the degrees of freedom as described in the supplementary materials of the paper. So after differentiating the log likelihood again I found a slightly different formula which passes the test nicely.  

@mzwiessele , there is one thing I think is important to fix at this point. The degrees of freedom parameter has a constraint, it should be greater than two. What would be the best option to achieve this? I tried logistic (]2, a big number]) but oddly, the optimization ends on the big number with no exception. I also tried LogexpClipped (as it has a lower attribute), but that attribute doesn't seem to be used in the code? Consequently, calling `randomize()` often resulted in values smaller than two. Finally, I simply took Logexp and add +2 in the code when using the param to shift the domain, but it doesn't seem appropriate to keep it like this.

### DIFF
--- a/GPy/inference/latent_function_inference/__init__.py
+++ b/GPy/inference/latent_function_inference/__init__.py
@@ -62,6 +62,7 @@ class InferenceMethodList(LatentFunctionInference, list):
             self.append(inf)
 
 from .exact_gaussian_inference import ExactGaussianInference
+from .exact_studentt_inference import ExactStudentTInference
 from .laplace import Laplace,LaplaceBlock
 from GPy.inference.latent_function_inference.var_dtc import VarDTC
 from .expectation_propagation import EP, EPDTC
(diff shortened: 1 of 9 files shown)


avehtari:
Based on the experiences by Stan development team
 -  logistic (]2, a big number])  is not recommended unless "a big number" is really a justified hard constraint (e.g. probabilities have to be between 0 and 1), and the performance of the inference can be sensitive to a specific value of "a big number". The fact that optimization ends to the 'a big number' indicates also that you are missing a Jacobian of the transformation.
 - log(nu-2) is just fine and used in Stan when specifying variable, e.g., 'real<lower=2> nu;' Of course it would be good to have this as a constraint with a parameter, and not something you need to code in the specific place.

Aki

PR #526, opened 2017-07-27
added extended version of MLP function

msbauer:
This is an extended version of the MLP mean function which supports multiple layers and several activation functions

### DIFF
--- a/GPy/mappings/__init__.py
+++ b/GPy/mappings/__init__.py
@@ -4,6 +4,7 @@
 from .kernel import Kernel
 from .linear import Linear
 from .mlp import MLP
+from .mlpext import MLPext
 from .additive import Additive
 from .compound import Compound
 from .constant import Constant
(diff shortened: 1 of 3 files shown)


zhenwendai:
It seems the unit-test fails because gradient check does not pass. Could you have a look into it?

msbauer:
The unittests for the new code run fine on my local machine with python 2.7. Rerunning the tests on the server, they now pass partly. I'm not sure where this problem comes from. Do you have any suggestions how to diagnose what might be going wrong?

zhenwendai:
Matthias, could you fix the randomness in the test case and update the pull request? Thanks!

msbauer:
Zhenwen, sorry for the long wait. I have fixed the seed and the checks now run except for the codecov/changes where I don't understand the coverage changes

zhenwendai:
Cool! Thanks, Matthias!

PR #527, opened 2017-07-27
Ll surv

adhaka:


adhaka:
Hii @alansaul , This pull request contains files for loglogistic and weibull likelihood along with the test cases, the loglogistic likelihood was written by you, i just included it with the pull request here, with the loggaussian one, I still need to change the parameterisation from p(y|f) to exp(y|exp(f)), anyways you could have a look and try to run this code on your machine. I will include the survival-notebook as well shortly.

zhenwendai:
@alansaul, could you have a look at this pull-request? It has been around for a while. I am not an expert about non-Gaussian likelihood.

alansaul:
Looks great to me! Since it is being tested in the same manner as the other likelihoods it should be fine (for Laplace at least). If the EP refactor is ready it would be cool to check whether it works with that as well (using quadrature rather than deriving the site terms for now would be reasonable). Having said that, I am happy to accept it as it is though with the caveat it may not work with EP. @adhaka I don't suppose you've had a chance to check it with EP have you?

zhenwendai:
@adhaka, could you update the pull request with devel branch after merging in the EP refactor branch.

adhaka:
Hi Zhenwen, 

Yeah sure. Thanks so much for merging the requests and following up on this :+1:  :)

adhaka:
Hii @alansaul , Thanks for the comments, I tried it a few times, and it was working with the new ep-refactored code, but have to work this out more rigourosly. If any fixes are required, I will make a new request for it. :+1:

alansaul:
That sounds great, let us know if there is anything we can do to help.

PR #545, opened 2017-09-11
Basic framework for serializing GPy models

pgmoren:


alansaul:
That's amazing Pablo, we definitely need this!

Is there some general instructions for how we add serialisation to our own classes, there seems to be a few caveats about how to build the dictionary that is subsequently serialised?
It appears that you usually choose to convert to a list of variables:

```
input_dict["variance"] = self.variance.values.tolist()
```
or an integer, rather than a numpy array - is that a necessity for example? You mention you assume Y_metadata is serialisable, it is often a dict with arrays inside, for example:

```
censored = np.zeros_like(self.Y)
random_inds = np.random.choice(self.N, int(self.N / 2), replace=True)
censored[random_inds] = 1
Y_metadata = {'censored': censored}
```

would this cause issues? I presume I would need to write a little wrapper that converts the arrays within the Y_metadata dict into lists, and then converts it back in to_dict and from_dict respectively?

pgmoren:
No, unfortunately I have not done any documentation, mostly because I wanted to push something quickly before this week, and because I wanted to get more feedback before working on the documentation. I have written below a rough description of what I have done. Any suggestion or recommendation is welcomed.

I have defined 4 methods (to_dict(), _to_dict(), from_dict() and _from_dict()) for the different components of a model. More specifically for the classes Likelihood, LatentFunctionInference, Kernel, Mapping, _Norm, GPTransformation(and derived classes).

And for the models themselves. In particular, for the class Model (and derived classes).

These methods are used to:

Convert to a serializable JSON dict
——————————————————

to_dict(): Converts the object to a dictionary that can subsequently be serialized to a JSON file. This dictionary needs a field “class” that contains a string indicating the class. I convert all the numpy arrays to lists because they are not JSON serializable. A more elegant solution I guess, would be to write a serializer for numpy arrays instead of using the to_list() and np.array() everywhere. In the base classes (Likelihood, LatentFunctionInference, Kernel, Mapping, _Norm, GPTransformation) the to_dict() raises a NotImplementedError. This is the function that you should implement  in your class derived from any of the previous based classes.

_to_dict(): It is an auxiliary function that it is used by to_dict() to get the dictionary. It is normally implemented in those classes that cannot be serialized, and it is called by the derived (serializable) class to get a dictionary with the fields of the (non serializable) parent class. 

Build the object given the serializable JSON dict
————————————————————————

from_dict(): It is implemented in the base class (e.g. Kern() for the kernels.) It extracts the class from the “class” field, and calls the _from_dict() method of that class. It defines the entry point for all the derived classes, and generally you should not need to define it in your derived classes.

_from_dict(): It is implemented in the base class. The implementation in the base class just calls the constructor of the class with parameters given by the remaining fields of the dictionary. This works directly for all cases in which the constructor has the same parameters as the ones saved in the dictionary. Otherwise, it is necessary to implement an additional _from_dict() for the class, that performs the additional logic.

In addition to these functions, I have implemented 3 methods (save_model(), _save_model(),  load_model()) for the class Model().

This functions are used to:

Save the model to a JSON file
—————————————————————————

save(): Calls _save() in those models that are serializable. In the base class it raises a NonImplementedError

_save(): Implemented only in only in the base class. It contains the actual logic. Most likely, you do not need to implement it in your class.

Load the model to a JSON file
—————————————————————————

load(): Implemented only in only in the base class. It contains the actual logic. Most likely, you do not need to implement it in your class.

Hopes this helps a little bit. It there is an interest I can put together a notebook with an example.

PR #549, opened 2017-09-12
fix: slight modification to MLP mapping to reduce potential for numpy…

christomaszewski:
… overflows and unnecessary computation. np.square(np.cosh(layer1)) was causing a numpy overflow error for some inputs. Since we have already computed the activations earlier in these functions, we can just make use of them directly instead of evaluating cosh() on the out output for the first layer (layer1). This modification eliminated the overflow for my data and passed all relevant tests.

### DIFF
--- a/GPy/mappings/mlp.py
+++ b/GPy/mappings/mlp.py
@@ -35,7 +35,7 @@ class MLP(Mapping):
 
         # Backpropagation to hidden layer.
         dL_dact = np.dot(dL_dF, self.W2.T)
-        dL_dlayer1 = dL_dact * (1 - np.power(activations, 2))
+        dL_dlayer1 = dL_dact * (1 - np.square(activations))
 
         # Finally, evaluate the first-layer gradients.
         self.W1.gradient = np.dot(X.T,dL_dlayer1)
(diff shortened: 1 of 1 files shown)


GPy/mappings/mlp.py
```diff
@@ -35,7 +35,7 @@ def update_gradients(self, dL_dF, X):
 
         # Backpropagation to hidden layer.
         dL_dact = np.dot(dL_dF, self.W2.T)
-        dL_dlayer1 = dL_dact / np.square(np.cosh(layer1))
+        dL_dlayer1 = dL_dact * (1 - np.power(activations, 2))
```
zhenwendai: Please use np.square instead of np.power(..., 2). In this case, np.square is more efficient.

GPy/mappings/mlp.py
```diff
@@ -47,7 +47,7 @@ def gradients_X(self, dL_dF, X):
...
```
zhenwendai: Please use np.square instead of np.power(..., 2). In this case, np.square is more efficient.

GPy/mappings/mlp.py
zhenwendai: Please use np.square instead of np.power(..., 2). In this case, np.square is more efficient.

zhenwendai (changes requested):

zhenwendai (approved):

Issue #266, opened 2015-10-13
Installation Error on Windows


vote539:
I am trying to upgade GPy because of issue #252.  I ran the following command, and got the error shown below.

```
PS C:\Users\me> pip install --upgrade https://github.com/SheffieldML/GPy/archive/master.zip

Collecting https://github.com/SheffieldML/GPy/archive/master.zip
  Downloading https://github.com/SheffieldML/GPy/archive/master.zip (1.5MB)
    100% |################################| 1.5MB 239kB/s
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "C:\Users\me\AppData\Local\Temp\pip-oikxnssj-build\setup.py", line 18, in read_to_rst
        f.write(pypandoc.convert('README.md', 'rst'))
      File "C:\WinPython\python-3.4.3.amd64\lib\encodings\cp1252.py", line 19, in encode
        return codecs.charmap_encode(input,self.errors,encoding_table)[0]
    UnicodeEncodeError: 'charmap' codec can't encode character '\u0292' in position 1732: character maps to <undefined>

    During handling of the above exception, another exception occurred:

    Traceback (most recent call last):
      File "<string>", line 20, in <module>
      File "C:\Users\me\AppData\Local\Temp\pip-oikxnssj-build\setup.py", line 88, in <module>
        long_description=read_to_rst('README.md'),
      File "C:\Users\me\AppData\Local\Temp\pip-oikxnssj-build\setup.py", line 20, in read_to_rst
        return read(fname)
      File "C:\Users\me\AppData\Local\Temp\pip-oikxnssj-build\setup.py", line 11, in read
        return open(os.path.join(os.path.dirname(__file__), fname)).read()
      File "C:\WinPython\python-3.4.3.amd64\lib\encodings\cp1252.py", line 23, in decode
        return codecs.charmap_decode(input,self.errors,decoding_table)[0]
    UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 2241: character maps to <undefined>

    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in C:\Users\me\AppData\Local\Temp\pip-oikxnssj-build
```

I'm running Python 3.4 on Windows 10.

Any ideas?

mzwiessele:
Yes use the devel branch:

pip install --upgrade https://github.com/SheffieldML/GPy/archive/devel.zip

vote539:
[...]

Issue #458, opened 2016-11-22
Issue when calling gp.predictive_gradients()


blaahhrrgg:
I am running Python v2.7.12 using Anaconda 64bit on Windows 7. I installed GPy using 'pip install GPy' to get version 1.5.6 which I have been using for a while on small problems without any issues. More, recently I have wanted to make use of the predictive_gradients() function but I have run into some issues. 

Firstly, when I import GPy I get the following warning: "warning in stationary: failed to import cython module: falling back to numpy" which I have not thought too much of in the past since everything else seemed to work okay. Secondly, when I try and run the following code,

```
import numpy as np
import GPy as gp

k = gp.kern.RBF(input_dim=1, variance=1., lengthscale=1.)

X = np.random.uniform(-3.,3., (20,1))
Y = np.sin(X) + np.random.randn(20,1)*0.05

m = gp.models.GPRegression(X, Y, kernel=k)

grad = m.predictive_gradients(X)
```

I get the following error,

```
NameError                                 Traceback (most recent call last)
<ipython-input-10-964a0d158030> in <module>()
----> 1 m.predictive_gradients(X)

C:\Program Files\Anaconda2\lib\site-packages\GPy\core\gp.pyc in predictive_gradi
ents(self, Xnew, kern)
    335
    336         for i in range(self.output_dim):
--> 337             mean_jac[:,:,i] = kern.gradients_X(self.posterior.woodbury_v
ector[:,i:i+1].T, Xnew, self._predictive_variable)
    338
    339         # gradients wrt the diagonal part k_{xx}

C:\Program Files\Anaconda2\lib\site-packages\GPy\kern\src\kernel_slice_operation
s.pyc in wrap(self, dL_dK, X, X2)
    116     def wrap(self, dL_dK, X, X2=None):
    117         with _Slice_wrap(self, X, X2) as s:
--> 118             ret = s.handle_return_array(f(self, dL_dK, s.X, s.X2))
    119         return ret
    120     return wrap

C:\Program Files\Anaconda2\lib\site-packages\GPy\kern\src\stationary.pyc in grad
ients_X(self, dL_dK, X, X2)
    234         """
    235         if config.getboolean('cython', 'working'):
--> 236             return self._gradients_X_cython(dL_dK, X, X2)
    237         else:
    238             return self._gradients_X_pure(dL_dK, X, X2)

C:\Program Files\Anaconda2\lib\site-packages\GPy\kern\src\stationary.pyc in _gra
[...]

PR #565, opened 2017-10-23
fix: dev: cython import errors

jayanthkoushik:
This commit fixes issues observed in Windows where some
cython modules are successfully imported, and some are not.
This causes the global config cython.working to be inconsistent,
which causes import errors when unavailable cython modules
are tried to be imported (example https://github.com/SheffieldML/GPy/issues/266).
This commit uses a separate flag for each module to fix the issue.

### DIFF
--- a/GPy/kern/src/coregionalize.py
+++ b/GPy/kern/src/coregionalize.py
@@ -9,10 +9,10 @@ from ...util.config import config # for assesing whether to use cython
 
 try:
     from . import coregionalize_cython
-    cython_coregionalize_working = True
+    use_coregionalize_cython = config.getboolean('cython', 'working')
 except ImportError:
     print('warning in coregionalize: failed to import cython module: falling back to numpy')
-    cython_coregionalize_working = False
+    use_coregionalize_cython = False
 
 
 class Coregionalize(Kern):
@@ -64,7 +64,7 @@ class Coregionalize(Kern):
         self.B = np.dot(self.W, self.W.T) + np.diag(self.kappa)
 
     def K(self, X, X2=None):
-        if cython_coregionalize_working and config.getboolean('cython', 'working'):
+        if use_coregionalize_cython:
             return self._K_cython(X, X2)
         else:
             return self._K_numpy(X, X2)
@@ -95,7 +95,7 @@ class Coregionalize(Kern):
             index2 = np.asarray(X2, dtype=np.int)
 
         #attempt to use cython for a nasty double indexing loop: fall back to numpy
-        if cython_coregionalize_working and config.getboolean('cython', 'working'):
+        if use_coregionalize_cython:
             dL_dK_small = self._gradient_reduce_cython(dL_dK, index, index2)
         else:
             dL_dK_small = self._gradient_reduce_numpy(dL_dK, index, index2)
--- a/GPy/kern/src/coregionalize.py
+++ b/GPy/kern/src/coregionalize.py
@@ -6,11 +6,14 @@ import numpy as np
 from ...core.parameterization import Param
 from paramz.transformations import Logexp
 from ...util.config import config # for assesing whether to use cython
+
 try:
     from . import coregionalize_cython
-    config.set('cython', 'working', 'True')
+    cython_coregionalize_working = True
 except ImportError:
-    config.set('cython', 'working', 'False')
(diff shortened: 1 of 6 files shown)


jayanthkoushik:
Sorry, the example issue is actually https://github.com/SheffieldML/GPy/issues/458.

zhenwendai:
Thanks a lot for identifying the issue and providing a fix. I notice that the fix will always take the cython if available for coregionalization, which cannot be controlled via the config file. Could you create a new entry for cython of coregionalization in the control file and follow the pattern of the existing cython control?

jayanthkoushik:
Hi, could you clarify what you mean by that? In coregionalize.py, I do check for the conda available property from config.

alansaul:
I think @zhenwendai is suggesting doing something like:

```
try:
    from . import coregionalize_cython
    use_cython = config.getboolean('cython', 'working')
except ImportError:
    use_cython = False
```

and then replace (throughout):
```
if cython_coregionalize_working and config.getboolean('cython', 'working'):
```
with
```
if use_cython:
```

would be a little cleaner, unless I'm missing some reason why we shouldn't do this (seems unlikely the config would change whilst running).

@jayanthkoushik is right, at the moment the current code automatically uses cython if it is available, through the lines:
```
try:
    from . import coregionalize_cython
    config.set('cython', 'working', 'True')
```
this in fact ignores the config file setting, which should perhaps be alternatively named 'try_use_cython', though of course it isn't worth changing. I think @lionfish0 ran into this issue recently when he was trying to debug some coregionalization model without using the cython code.

jayanthkoushik:
I think that makes sense. I've modified the code accordingly; I left the test scripts as they were though, because I think it makes sense to test cython if possible, regardless of the setting (which as you point out, should represent preference rather than availability). I also used separate flags for each module, since that allows programatically checking for the availability of particular cython modules by importing the respective flags.

mzwiessele:
We should add the test cases for the windows runs, so that the cython modules are being imported. This would greatly increase the confidence in the automatic building process.

mzwiessele:
As soon as the tests are in place, we can merge :)

jayanthkoushik:
Sorry, I don't quite follow. Which test scripts need to be modified?

mzwiessele:
I mean to put in a dynamic test, which tests both versions of the code, the numpy as well as cython implementations. I believe that has been tested in the kernel_tests.py before

https://github.com/SheffieldML/GPy/blob/34a5e7ed700fb1ccf97b803bc7d2676c03e23e4b/GPy/testing/kernel_tests.py#L632

but somehow fails to work now

https://coveralls.io/builds/14246027/source?filename=GPy%2Fkern%2Fsrc%2Fcoregionalize.py#L111

It has been cotrolled by using the config and running both versions of the code and then comparing the results to be the same.

Does that make sense? I think I over complicated things, I just mean that the tests should be testing both versions by making use of the config. That both tests that the config works and that both versions of the code return the same results.

mzwiessele:
Btw. good spotting, thank you very much for contributing :)!

jayanthkoushik:
Ok, I think that makes sense. I changed the test script to always use cython if possible, and ignore the global config. Using the config as well should fix the issue right?

mzwiessele:
I think so, yes :) the test itself is using the global config to switch the versions, so it should now cover both, numpy and cython code in the test. Let’s see what the CI says when it comes back.

Thanks again!

mzwiessele:
Have you committed the version where both cython and numpy are being tested?

jayanthkoushik:
No, will do it soon.

jayanthkoushik:
Committed - fingers crossed :)

jayanthkoushik:
Hmmm, that didn't seem to do it. I'll take a look.

jayanthkoushik:
Ok, I see the problem. In kernel_tests.py, the global config is changed within the test, which is not reflected since the global config is checked only on import.

jayanthkoushik:
I think the cython related coverage issues are resolved. The unexpected changes don't seem to be related to any code I've changed.

lawrennd:
@mzwiessele @jayanthkoushik is there any reason this one is still hanging? Looks like it got very close to being integrated.

jayanthkoushik:
Yeah I this it is ready to be integrated.

PR #572, opened 2017-11-14
Add mean function functionality to dtc inference method

marpulli:
Allows us to use mean functions with sparse GP regression. The changes are mainly adding mean_function as an input to required methods and then a small reordering of the maths to get dL_dm out.

### DIFF
--- a/GPy/testing/inference_tests.py
+++ b/GPy/testing/inference_tests.py
@@ -49,6 +49,7 @@ class InferenceXTestCase(unittest.TestCase):
         m.optimize()
         x, mi = m.infer_newX(m.Y, optimize=True)
         np.testing.assert_array_almost_equal(m.X, mi.X, decimal=2)
+
 class InferenceGPEP(unittest.TestCase):
 
     def genData(self):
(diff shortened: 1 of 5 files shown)


alansaul:
This looks great thanks for the contribution. Sorry the tests were previously failing that was our fault for not keeping builds up to date. Merging.

GPy/testing/inference_tests.py
```diff
@@ -132,6 +133,16 @@ def test_inference_EP_non_classification(self):
...
+class VarDtcTest(unittest.TestCase):
+
+    def test_var_dtc_inference_with_mean(self):
+        """ Check dL_dm in var_dtc is calculated correctly"""
+        np.random.seed(1)
+        x = np.linspace(0.,2*np.pi,100)[:,None]
+        y = -np.cos(x)+np.random.randn(*x.shape)*0.3+1
+        m = GPy.models.SparseGPRegression(x,y, mean_function=GPy.mappings.Linear(input_dim=1, output_dim=1))
+        self.assertTrue(m.checkgrad())
+
```
zhenwendai, line 145: Can we have a unittest that verify the prediction of sparse GP with mean function?

alansaul: Hmm, good point I hadn't thought of that. Unless I'm missing them we don't have such tests for the vanillia GP prediction function either (which we should)? Or do you just mean testing that the prediction function doesn't throw an exception? As in:
https://github.com/SheffieldML/GPy/blob/83c3ed44e28b5996e262fcaba18d612d63d6e274/GPy/testing/gp_tests.py#L96

PR #574, opened 2017-11-14
Fixing scipy=1.0.0 incompatibility of lyapunov discovered in PR #573

alansaul:
As discussed in PR #573 scipy bug, scipy/scipy#8155, caused failing import for users of scipy=1.0.0 this is a simply work around to handle this edge-case.

P.s. we should delete kern/_src unless it is used for something I am unaware of.

### DIFF
--- a/GPy/kern/_src/sde_stationary.py
+++ b/GPy/kern/_src/sde_stationary.py
@@ -9,6 +9,10 @@ from .stationary import RatQuad
 
 import numpy as np
 import scipy as sp
+try:
+    from scipy.linalg import solve_continuous_lyapunov as lyap
+except ImportError:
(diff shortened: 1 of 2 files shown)


alansaul:
Strange, am I crazy or is it saying that those 4 lines changed the coverage by ~ -3.2%?

It seems almost all to do with prediction/plotting, could the change to scipy==1.0 be also making a difference here for some reason? @mzwiessele, any idea? Something to do with the plotting tests?

The three lines can't feasibly be tested without modifying travis to test both scipy < 1.0 and scipy == 1.0 which won't be represented in the coverage.

alansaul:
Coverage issue should be resolved by PR #575, merging.

PR #575, opened 2017-11-15
Trying to fix tests for Matplotlib plotting issue

alansaul:
Seeing whether the matplotlib backend import is the plotting test issue that reduces coverage.


alansaul:
Merging this. Although it 'breaks' the build, this is only due to miniconda using a more up to date matplotlib version which removed ImageComparisonFailure, this isn't even used in GPy.

PR #578, opened 2017-11-20
Fix EP for non-zero mean GP priors (binary classification)

pgmoren:


GPy/inference/latent_function_inference/expectation_propagation.py
```diff
@@ -81,35 +82,40 @@ def __init__(self, mu, Sigma, L=None):
         Sigma_diag = np.diag(self.Sigma)
         super(posteriorParams, self).__init__(mu, Sigma_diag)
 
-    def _update_rank1(self, delta_tau, ga_approx, i):
-        ci = delta_tau/(1.+ delta_tau*self.Sigma_diag[i])
-        DSYR(self.Sigma, self.Sigma[:,i].copy(), -ci)
-        self.mu = np.dot(self.Sigma, ga_approx.v)
+    def _update_rank1(self, mean_prior, delta_tau, delta_v, ga_approx, i):
```
alansaul: Does mean_prior need to be passed in if it is never used, is its effect somehow ? Maybe this is more general for other algorithms that are based off this posterior?

GPy/inference/latent_function_inference/expectation_propagation.py
```diff
@@ -244,48 +256,46 @@ def expectation_propagation(self, K, Y, likelihood, Y_metadata):
...
             assert self.ga_approx_old.v.size == num_data, "data size mis-match: did you change the data? try resetting!"
             ga_approx = gaussianApproximation(self.ga_approx_old.v, self.ga_approx_old.tau)
-            post_params = posteriorParams._recompute(K, ga_approx)
+            post_params = posteriorParams._recompute(mean_prior, K, ga_approx)
             diag.add(post_params.Sigma, 1e-7)
             # TODO: Check the log-marginal under both conditions and choose the best one
         return (ga_approx, post_params)
 
-    def _local_updates(self, num_data, cav_params, post_params, marg_moments, ga_approx, likelihood, Y, Y_metadata, update_order=None):
+    def _local_updates(self, mean_prior, K, num_data, cav_params, post_params, marg_moments, ga_approx, likelihood, Y, Y_metadata, update_order=None):
```
alansaul: Given the other comment, it's not clear to me how mean_prior is being used when doing local updates? Does it belong in here as well?

alansaul (commented):
Great this is totally needed! Just a couple of questions where its not quite clear why the mean_prior is passed in. Maybe an test on the posterior might be useful? Something really simple like a long length-scale GP, Gaussian likelihood, constant mean mean function, check that the posterior is approximately constant with that offset? We should probably make something like this a general test for all inference methods that have mean functions implemented just as a sanity check.

GPy/inference/latent_function_inference/expectation_propagation.py
pgmoren: It is not, and also K is not used. I was using them for debugging and somehow forgot to delete them. I will create a new pull request after amending this. Thanks a lot!

pgmoren:
Hi, I have amended the commit addressing your comments respect to the unused variables. Thanks :) I will think more about new unit test for EP and create a new pull request.

zhenwendai:
@alansaul, are you OK with merging this pull request?

alansaul:
Absolutely, please go ahead.

At some point we should try figure out a more robust way of testing the predictions of all the models with mean functions (GP, sparse GP, EP, Laplace - when I get it done) to make sure they are doing something sensible but lets leave it for now. There is a test included in this PR that checks that it at least isn't throwing an error, which is all we are testing for the other models with mean functions implemented.

zhenwendai:
Testing prediction with mean function is particularly hard, because GP could overwrite the mean function to fit data. 

One way that we could do is probably to compare with pre-recorded numbers.

Issue #81, opened 2013-09-20
Automatically converting lists,matrices to arrays?


jamesmcm:
When I first used GPy, I had a confusing problem where I was passing the labels as a list rather than a numpy array, this wasn't obvious (in that case with util.plot_latent ).

We could add code to automatically convert from other data types to the numpy arrays? Would this be recommended?

PR #579, opened 2017-11-23
Improve the documentation for LVMOGP

zhenwendai:


GPy/models/gp_multiout_regression.py
```diff
@@ -13,26 +13,28 @@
...
-    :type W_rank: integer
-    :param kernel_name: name of the kernel
-    :type kernel_name: string
+    Gaussian Process model for multi-output regression without missing data
+
+    This is an implementation of Latent Variable Multiple Output Gaussian Processes (LVMOGP) in [Dai et al. 2017].
+
+    Zhenwen Dai, Mauricio A. Álvarez and Neil D. Lawrence. Efficient Modeling of Latent Information in Supervised Learning using Gaussian Processes. In NIPS, 2017.
+
+    :param X: input observations. Numpy.ndarray
```
alansaul: For consistency with the rest of the doccing I've been doing, it would be good to specify the types like;
'''
:type X: np.ndarray (num_data x input_dim)
'''

etc. so that users know what shapes things should be (perhaps obvious for some, but less obvious for things like Z_row and qU_var_c_W_dim).

alansaul:
Looks great, thanks Zhenwen. Ideally you could add types to the input parameters otherwise it's good to be merged as far as I'm concerned.

I don't suppose there is a simplified notebook from the paper that we could use to show off some of the models features and how to use it is there? I'm trying to tidy up the rest of them and it would be great to have something for all the major models inside GPy.

zhenwendai:
I don't know whether we want to include the types for all the input arguments. Argument types do not contain much information, but make the documentation significantly longer, but let me know if you think we should include them.

It is a good idea to collect the notebooks for all the models. I plan to make a notebook demonstrating the usage of these two models. Where do we host them? I guess we will use the notebook repo?

mzwiessele:
I would say if the type is not obvious, or a class defined in GPy, the type should definitely be in the documentation, so the linking can be used and you can click through in the browser. 

Otherwise, great work! I am glad GPy goes forward again :)

The notebooks will be the website, as discussed earlier. So that everything GPy is in one place.

alansaul:
I agree there is positives and negatives to having type of inputs included (larger proportion of code is doc strings, but each is more informative).

Personally I would prefer to have the types specified so that new-comers can understand the code more easily (note these are only public facing functions that are being doc-stringed) and that is what I have been doing whilst documenting the code. I prefer the consistency of everything being documented as well, and that in some cases they help a lot. I can remove them if there is strong feelings that they shouldn't included for brevity of the code though.

Perhaps when I've finished (I'm obviously doing other things in the doc strings so its not a huge issue adding the types as I go along) you can take a look and decide whether they are useful or a hinderance.

Feel free to miss them out for this model for now, we can always add them later. Of course if they are only simple types, there is a middle ground of just doing:

```
:param float X_row:
```

Which doesn't really increase the amount of documentation but gives *some* information about the type. Having said that, I have no idea whether X_row is supposed to be a float or a certain shaped ndarray - I'm guessing np.ndarray (num_inputs x input_dim) ? - and that is kind of why I think types could help.

Notebooks for now will be hosted in SheffieldML/notebook as before, I am arranging notebooks of different types of models into different folders to help Tania with her notebook hosting code. Don't worry about this for now though just shove it in I will sort it when the new website it ready, I'll document the folder structure in due course when it is rolled out.

zhenwendai:
> Which doesn't really increase the amount of documentation but gives some information about the type. Having said that, I have no idea whether X_row is supposed to be a float or a certain shaped ndarray - I'm guessing np.ndarray (num_inputs x input_dim) ? - and that is kind of why I think types could help.

More information is definite helpful. I just wonder whether it is in the format of argument types, because it alone does not tell much. In this example, it contains not only a type but also a shape. It is not easy to specify a shape because it need contextual information such as what is num_inputs and what is input_dim, which is not explained at all. I think the best way is to give some examples, but it is time consuming to make and a bit overlapping with the role of notebooks.

mzwiessele:
I would like to put your attention to the old misunderstanding from #81, which was due to not specifying the proper input types. We decided back then, that the specification of types will reduce the amount of type checking immensly and relieve us of catching corner cases.

I think we should still hold on to these ideas, to reduce the load on us in the long run.

What do you think?

zhenwendai:
In this case, it makes sense to provide types. I will add types into the docstrings.

max-li-dd:
Hi @zhenwendai @alansaul @mzwiessele,

Thanks for the great software! I was wondering if there is notebook for the examples of LVMOGP?

Thanks,
Max

PR #585, opened 2017-12-18
modify the MLP kernel equation

YoshikawaMasashi:
Modify the MLP kernel equation in the document.


zhenwendai:
Thanks for the correction!

PR #589, opened 2018-01-05
Implemented utility function to compute covariance between points in GP Model

apaleyes:
This change is not directly used in the library, but is needed for the entropy search acquisition implementation in GPyOpt

### DIFF
--- a/GPy/core/gp.py
+++ b/GPy/core/gp.py
@@ -8,7 +8,6 @@ from .mapping import Mapping
 from .. import likelihoods
 from .. import kern
 from ..inference.latent_function_inference import exact_gaussian_inference, expectation_propagation
-from ..util.linalg import dtrtrs
 from ..util.normalizer import Standardize
 from paramz import ObsAr
 
@@ -680,23 +679,11 @@ class GP(Model):
         mu_star, var_star = self._raw_predict(x_test)
         return self.likelihood.log_predictive_density_sampling(y_test, mu_star, var_star, Y_metadata=Y_metadata, num_samples=num_samples)
 
-    def posterior_covariance(self, X1, X2):
+    def posterior_covariance_between_points(self, X1, X2):
         """
         Computes the posterior covariance between points.
 
         :param X1: some input observations
         :param X2: other input observations
         """
-        # ndim == 3 is a model for missing data
-        if self.posterior.woodbury_chol.ndim != 2:
-            raise RuntimeError("This method does not support posterior for missing data models")
-
-        Kx1 = self.kern.K(self.X, X1)
-        Kx2 = self.kern.K(self.X, X2)
-        K12 = self.kern.K(X1, X2)
-
-        tmp1 = dtrtrs(self.posterior.woodbury_chol, Kx1)[0]
-        tmp2 = dtrtrs(self.posterior.woodbury_chol, Kx2)[0]
-        var = K12 - tmp1.T.dot(tmp2)
-
-        return var
+        return self.posterior.covariance_between_points(self.kern, self.X, X1, X2)
--- a/GPy/core/gp.py
+++ b/GPy/core/gp.py
@@ -8,6 +8,7 @@ from .mapping import Mapping
 from .. import likelihoods
 from .. import kern
 from ..inference.latent_function_inference import exact_gaussian_inference, expectation_propagation
+from ..util.linalg import dtrtrs
 from ..util.normalizer import Standardize
 from paramz import ObsAr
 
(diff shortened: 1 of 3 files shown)


alansaul:
I was wondering, perhaps this would be more well suited to live in `GPy.inference.latent_function_inference.posterior`, perhaps alongside this:

https://github.com/SheffieldML/GPy/blob/97639ff2a6ee94f96e4840a1687616a099e3c7fe/GPy/inference/latent_function_inference/posterior.py#L90

The name `posterior_covariance` is a bit confusing to me though, as m.posterior.covariance is what I would think of as the posterior covariance, i.e. the covariance of the posterior. Can you think of a more suitable name, unless I'm mistaken?

mzwiessele:
I agree with Alan, it should go into the posterior covariance. Then it will also apply for missing data, right?

We could give the covariance function two optional arguments X1 and X2 and depending on the availibility of those, we compute the right covariance.

apaleyes:
I tried to move it to posterior class before, but it became too much dependent on the kernel calculations provided by the model, so result looked super weird. Thus I left it in the model.

As for the name, the idea is that the line

    model.posterior_covariance(X1, X2)

would read as "model, give me a posterior covariance between points X1 and X2". This looks intuitive to me, but to make it more explicit we can call the method "posterior_covariance_between_points". Does that sound any better?

mzwiessele:
Can you explain the issue you are having in greater detail? It should be the same kernel you compute on, with the same parameters if I am not mistaken?

apaleyes:
@mzwiessele sure. Here is the bit:

        Kx1 = self.kern.K(self.X, X1)
        Kx2 = self.kern.K(self.X, X2)
        K12 = self.kern.K(X1, X2)

Here `self` is the model. If `posterior_covariance` was a method in the posterior class, it would need to get the results of these computation somehow. There are two options:

1. Precompute Kx1, Kx2 and K12 outside and pass them in as the parameters. The problem here is that it is very hard to describe these parameters to make a good method in Posterior class. I just don't see how they make sense outside of this particular method.
2. Pass `self.X` and `self.kern` to Posterior, so that it computes everything itself. That means that we have a method in GP class calling a method in Posterior class, which seems superfluous to me.

Besides, there is also a thing about how this is used. In GPyOpt, or any other client code that would call this method, user would normally only have a reference to the model object. One of course could call `model.posterior.covariance_between_points`, but this looks like interface violation, because it makes some assumptions about how GP model class is implemented. So I believe we need a method for that in GP class anyway.

Hopefully I did not explain that too clumsily. Does that any sense at all?

mzwiessele:
To me it is another predict method, which predicts the covariance between
given points. Therefore, it should live in the posterior object and the
kernel is just a parameter. This means you can predict with another kernel
in case you want that using the posterior learnt by the GP. The way of
calling m.covariance.postiorior(kern, X1, X2) seems fine to me. The
prediction methods on the GP class are just convenience, as the posterior
is actually the actor doing predictions.

2018-01-05 16:16 GMT+01:00 Andrei Paleyes <notifications@github.com>:

apaleyes:
Thanks, I am finally convinced Posterior is where this function should live. Updated.

I did not move the tests anywhere because there seems to be no separate test file for Posterior, and current tests convey usage intention well.

alansaul:
Looks good to me. I would say, if you like (and to aid compatibility for other GPyOpt approximating function implementations, which is it's current use-case), it is fine to add a wrapper in GPy/core/gp.py which calls it with the defaults self.kern and self.X . We already have this for predict, etc. 

I totally see where you are coming from with above two arguments, passing self.X and self.kern is not as neat as we would like. We could additionally hold the self.X and self.kern that created K (and hence woodbury_chol) in the Posterior object, as the object should be recreated each time these change anyway, as K must change in response. Though obviously it is wise to keep it in only one place, the highest up object, i.e. gp.py. I think it's probably worth not changing things as they stand now though as it requires them to be stored twice which is not nice, and its not a huge head-ache passing them along as you have done - especially if you add the wrapper function for GPy/core/gp.py as proposed above.

Naming wise I think the function name is okay, but would it be more accurate to call it the predictive_covariance_between_two_points? If you feel the name is correct that is fine and I am happy with the merge.

apaleyes:
Hey Alan, the only reason I did not call it "posterior_..." is that the resulting code would look like:
"m.posterior.posterior_covariance_between_points", and this "posterior.posterior" does not look nice at all.

However this name makes sense for the wrapper in GPModel you suggested. I'll add it now

alansaul:
You're correct of course Andrei, please carry on.

apaleyes:
Wrapper is in, all checks pass.

alansaul:
I'm happy for this to be merged.

zhenwendai:
Cool, let's merge it.

PR #597, opened 2018-02-05
Allow calculation of full predictive covariance matrices with multipl…

marpulli:
This code didn't work before:

```
import GPy 
import numpy as np

x = np.random.rand(5, 1)
y = np.random.rand(5, 2)

model = GPy.models.GPRegression(x, y, normalizer=True)
model.predict(x, full_cov=True)
```

This PR adds "inverse_covarance" to normalizer so we know how to unscale the full covariance matrix. Behaviour for other cases should be unchanged.

### DIFF
--- a/GPy/core/gp.py
+++ b/GPy/core/gp.py
@@ -282,10 +282,12 @@ class GP(Model):
             mu += self.mean_function.f(Xnew)
         return mu, var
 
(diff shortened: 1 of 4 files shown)


mzwiessele:
A yes good point, thanks!

Issue #582, opened 2017-12-05
Plotting with SparseGPRegression


lionfish0:
I've found a couple of maybe bugs. Probably related so putting them in same issue.

1) Code example:
```
import GPy
import numpy as np
import matplotlib.pyplot as plt
%matplotlib inline
x = np.random.rand(100,3)
x[0:10,-1] = 0
x[10:20,-1] = 1
x[20:30,-1] = 2
x[30:50,-1] = 3
x[50:,-1] = 4
y = (x[:,0]+x[:,1]+x[:,-1])[:,None]

print(x.shape)
print(y.shape)
k = GPy.kern.RBF(2,ARD=True)**GPy.kern.Coregionalize(1,output_dim=5,rank=1)
#works:
#m = GPy.models.GPRegression(x, y, k)
#causes plot to fail:
m = GPy.models.SparseGPRegression(x, y, k)
m.plot(fixed_inputs=[(0,0),(2,4)])
```
Fails to plot, and excepts with:
```
(100, 3)
(100, 1)

---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-5-121686a4acbc> in <module>()
     18 #causes plot to fail:
     19 m = GPy.models.SparseGPRegression(x, y, k)
---> 20 m.plot(fixed_inputs=[(0,0),(2,4)])
     21 
     22 #removing the coregionalisation, we find we still can't plot...

~/anaconda3/lib/python3.6/site-packages/GPy/plotting/gpy_plot/gp_plots.py in plot(self, plot_limits, fixed_inputs, resolution, plot_raw, apply_link, which_data_ycols, which_data_rows, visible_dims, levels, samples, samples_likelihood, lower, upper, plot_data, plot_inducing, plot_density, predict_kw, projection, legend, **kwargs)
    338     plots = {}
    339     if hasattr(self, 'Z') and plot_inducing:
--> 340         plots.update(_plot_inducing(self, canvas, visible_dims, projection, 'Inducing'))
    341     if plot_data:
    342         plots.update(_plot_data(self, canvas, which_data_rows, which_data_ycols, free_dims, projection, "Data"))

~/anaconda3/lib/python3.6/site-packages/GPy/plotting/gpy_plot/data_plots.py in _plot_inducing(self, canvas, visible_dims, projection, label, **plot_kwargs)
    172 def _plot_inducing(self, canvas, visible_dims, projection, label, **plot_kwargs):
    173     if visible_dims is None:
[...]

PR #600, opened 2018-02-12
Fix visible dimensions for plotting inducing points

marpulli:
Fixes bugs with plotting sparse GPs. Fixes issue https://github.com/SheffieldML/GPy/issues/582


mzwiessele:
Good spotting! Thanks

PR #607, opened 2018-03-07
Add serialization functions for EPDTC

pgmoren:


zhenwendai:
This pull-request only contains changes to EP code, but the unit-tests fail in plotting test cases. @mzwiessele, can you have a look into the error in the unit-tests?

zhenwendai:
With manual investigation, we found that the failure of testcase was due to the upgrade of matplotlib 2.0 -> 2.2.

PR #613, opened 2018-03-21
Fixes the dimensions of the samples output

dtorrejo:
The dimensions of the output should respect the number of the points from input X and not from the training data self.num_data


mzwiessele:
Good spotting! Could you change it to X.shape[0] to be consistent with the need for numpy arrays as inputs?

dtorrejo:
No problem. Made the change.

PR #648, opened 2018-07-02
Symmetric kernel

marpulli:
A kernel that enforces odd or even symmetry across a given plane.

### DIFF
--- a/GPy/kern/src/symmetric.py
+++ b/GPy/kern/src/symmetric.py
@@ -30,9 +30,13 @@ class Symmetric(Kern):
         k(x, x') \pm k(Ax, x') \pm k(x, Ax') + k(Ax, Ax')
 
     where k(x, x') is the kernel of g(x)
+
+    :param base_kernel: kernel to make symmetric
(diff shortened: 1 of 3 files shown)


GPy/kern/src/symmetric.py
```diff
@@ -0,0 +1,137 @@
...
+        f(x) = g(x) \pm g(Ax)
+
+    with kernel
+
+    .. math::
+
+        k(x, x') \pm k(Ax, x') \pm k(x, Ax') + k(Ax, Ax')
+
+    where k(x, x') is the kernel of g(x)
+    """
```
zhenwendai: I can get the definition of this kernel, but for a person who is less familiar with GP it would be confusing. Please expand the description with a bit more details, e.g. "\pm" is confusing, it is better to use the sign variable or something.

zhenwendai (changes requested):
The changes look very nice. Please expand a bit the class description.

zhenwendai (approved):

PR #655, opened 2018-07-24
Allow setup.py to be parsed without numpy

### LINKED ISSUE #653: Unable to parse setup.py without numpy installed
This means you can't `pip download` gpy, for example.

```
davids@davids:~/mord$ mkvirtualenv test
Using base prefix '/usr'
New python executable in /home/davids/.virtualenvs/test/bin/python3
Also creating executable in /home/davids/.virtualenvs/test/bin/python
Installing setuptools, pip, wheel...done.
(test) davids@davids:~/mord$ pip download gpy
Collecting gpy
  Downloading https://files.pythonhosted.org/packages/0f/0a/856e3d2990739ae4a4dc526a1987dbf31ae3310d051322df815147e76453/GPy-1.9.2.tar.gz (859kB)
    100% |████████████████████████████████| 860kB 4.8MB/s
  Saved ./GPy-1.9.2.tar.gz
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-download-qrakgoby/gpy/setup.py", line 41, in <module>
        import numpy as np
    ImportError: No module named 'numpy'

    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-download-qrakgoby/gpy/
```

I guess the best solution would be to move the import, and the definition of `ext_mods` into a `try` block. This way pip would be able to determine that `numpy` is a dependency, and can install it.

davidsmf:
If numpy isn't available, don't define ext_mods, pip will then determine numpy is required, install it, then call us again.

Fixes #653

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,6 @@ from __future__ import print_function
 import os
 import sys
 from setuptools import setup, Extension
-import numpy as np
 import codecs
 
 def read(fname):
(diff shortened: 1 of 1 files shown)


zhenwendai:
Cool! Thanks for the fix! Usually Linux/mac comes with a version of numpy, so we do not experience this problem.

PR #663, opened 2018-08-25
Added cython to setup_requires

KOLANICH:


mzwiessele:
Cython is  not a requirement for this package. Thank you very much for the suggesting 😃

PR #689, opened 2018-10-12
fix: basis functions gradients pushed up a level in hierarchy

mzwiessele:
This is to alleviate gradient computations a level up, only having to provide 
frac{d\phi}{d\theta}


ekalosak:
Closed due to inactivity.

Issue #692, opened 2018-10-18
Install fails - linux - compile issues - numpy import_array()


ajgpitch:
Hello,
I was unable to install on Linux. I tried `pip` and from source. There did not seem to be any error, just warnings relating to `numpy import_array()`. The console output and conda env list is copied below.

I have done a bit of searching and found a solution. I will make a PR of the proposed fix just now.

# Console output
```
(bayesopt) agp1@imapspc0049 ~/GitHub/ajgpitch/GPy $ python setup.py develop
running develop
running egg_info
writing GPy.egg-info/PKG-INFO
writing dependency_links to GPy.egg-info/dependency_links.txt
writing requirements to GPy.egg-info/requires.txt
writing top-level names to GPy.egg-info/top_level.txt
reading manifest file 'GPy.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
warning: no files found matching 'README.rst'
writing manifest file 'GPy.egg-info/SOURCES.txt'
running build_ext
building 'GPy.kern.src.stationary_cython' extension
gcc -pthread -B /home/agp1/anaconda3/envs/bayesopt/compiler_compat -Wl,--sysroot=/ -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -fPIC -I/home/agp1/anaconda3/envs/bayesopt/lib/python3.7/site-packages/numpy/core/include -I. -I/home/agp1/anaconda3/envs/bayesopt/include/python3.7m -c GPy/kern/src/stationary_cython.c -o build/temp.linux-x86_64-3.7/GPy/kern/src/stationary_cython.o -fopenmp -O3
In file included from /home/agp1/anaconda3/envs/bayesopt/lib/python3.7/site-packages/numpy/core/include/numpy/ndarraytypes.h:1823:0,
                 from /home/agp1/anaconda3/envs/bayesopt/lib/python3.7/site-packages/numpy/core/include/numpy/ndarrayobject.h:18,
                 from /home/agp1/anaconda3/envs/bayesopt/lib/python3.7/site-packages/numpy/core/include/numpy/arrayobject.h:4,
                 from GPy/kern/src/stationary_cython.c:242:
[...]

PR #695, opened 2018-10-18
extension sources changed to pyx

ajgpitch:
Although this seems a bit contradictory and confusing
https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html
on whether to specify `.pyx` or `.c` as the ext source 

However, if this this PR were implemented, then I could have worked around #692  installing from source with the '--force' option


ajgpitch:
I would also suggest that `.gitignore` should exclude any cython generated `.c` files. So that by default  installing from a clone cythonizes the `.pyx` files. Only include the cython generated `.c` files in the `bdist`.

PR #697, opened 2018-10-19
FIX: state_space_cython misc

ajgpitch:
add missing colons and fix indentation

PR #719, opened 2019-01-10
FIXES - for installing under Python 3.7

ajgpitch:
This is a combination of #693 #695 #697 . I think all three of these are required for successful install under Python 3.7 on any platform. The current `pip` and install from source fail on Python 3.7

I have tested locally with Python 2.7 and 3.7 on Linux Mint 19 and found this to be error free. I have made this PR to demonstrate that this combo solves the install and other issues, i.e. it should pass CI tests. 

I would suggest that the three PRs are merge separately, as this will be much tidier in the log. This PR can then be closed, as it contains nothing except whats in #693 #695 #697


mzwiessele:
Thank you so much!!!

PR #720, opened 2019-01-10
ENH: Untrack .c files generated by Cython

ajgpitch:
It is not necessary to store these files in Github. They should be generated locally after a clone. 
Tracking them leads to unnecessary commits, such as the last one in #719. 

Can also mask issues in underlying `.pyx` files. I assume this is how the issues fixed in #697 must have made it into the `devel` branch. Otherwise tests could not have passed.

`.c` files added to `.gitignore`
cythonised c files untracked

Any future C files which are not generated by Cython can be added using
```
git add -f
```
The two C files that fit that description are not untracked by this PR.


ekalosak:
This is great, thanks for improving the code sanitization 🙏 
Problem is it's not passing CI (the build expects these `.c` files).
Once that's fixed, this LGTM, good engineering.

ekalosak:
Crap, ok, so upon reflection - this is a great upkeep PR and I missed merging it and now it's out of date. If you update it I'd say let's just get it in. Otherwise this PR will hang here for a moment longer.

ekalosak:
The only thing blocking this (otherwise correct) PR is to update the CI so it builds - and therefore doesn't require - the `.c` files you've removed. It's the right move, just incomplete.

ekalosak:
Closing due to no activity & merge conflicts

PR #759, opened 2019-05-17
Deploy 1.9.7

zhenwendai:
This will enable Python 3.7 support.


mzwiessele (approved):
I bumped the version so that we can finalize the builds.
The build will be done automatically from deploy as always :)

Issue #728, opened 2019-01-30
Plotting Broken


tyler-abbot:
Plotting is broken with a dimension mismatch error. Here is a minimum working example:
```python
import GPy
import numpy as np

x = np.linspace(0.0, 1.0, 100)[:, np.newaxis]
y = np.linspace(0.0, 1.0, 100)[:, np.newaxis]

model = GPy.models.GPRegression(x, y)

fig = model.plot()
```
This issue was mentioned in #723. The user mentions that f2py is having some issue, but does not specify a solution, so I'm reopening the issue here.

himakarganti:
Here is what works & doesnt work:

WORKS:
CentOS Platform:
GPy - v 1.9.2 & Scipy - v 1.1.0

DOES NOT WORK:
OS X platform: 
GPy - v 1.9.6 & Scipy - v 1.2.0
GPY - v 1.9.2   &  Scipy - v 1.2.0

Here is my error:
    model1.plot()
  File "/usr/local/lib/python2.7/site-packages/GPy/plotting/gpy_plot/gp_plots.py", line 324, in plot
    helper_data = helper_for_plot_data(self, X, plot_limits, visible_dims, fixed_inputs, resolution)
  File "/usr/local/lib/python2.7/site-packages/GPy/plotting/gpy_plot/plot_util.py", line 121, in helper_for_plot_data
    Xgrid[:,free_dims] = Xnew
ValueError: shape mismatch: value array of shape (200,1,1) could not be broadcast to indexing result of shape (1,200).

palindromik:
This error could be related to the latest release of numpy (v 1.16.0 on Jan 14th: https://pypi.org/project/numpy/1.16.0/#history).

On Mac OS X, the given model.plot example fails when updating the numpy version while keeping other dependencies constant.
Works: GPy - 1.9.6 and numpy - **1.15.4**
Doesn't work: GPy - 1.9.6 and numpy - **1.16.0**

ThomasScragg:
I have just started working with GPy and while working through the tutorial examples the plotting problem arose. 
Using jupyter notebooks on two machines I have managed to get two working combinations:
SciLinux 6.2:         GPy 1.9.2, scipy 1.1.0 and numpy 1.15.4 - plotting works
Windows 10 Home (64 bit) :      GPy 1.9.6, scipy 1.2.0 and numpy 1.15.4 - plotting works.

kakila:
[...]

Issue #723, opened 2019-01-16
model.plot() not working in GPy 1.9.6


himakarganti:
The plotting function was working in the previous version GPY v 1.9.2; but after updating to the latest (1.9.6) I get the following error:

_**_model.plot()
  File "/usr/local/lib/python2.7/site-packages/GPy/plotting/gpy_plot/gp_plots.py", line 324, in plot
    helper_data = helper_for_plot_data(self, X, plot_limits, visible_dims, fixed_inputs, resolution)
  File "/usr/local/lib/python2.7/site-packages/GPy/plotting/gpy_plot/plot_util.py", line 121, in helper_for_plot_data
    Xgrid[:,free_dims] = Xnew
ValueError: shape mismatch: value array of shape (200,1,1) could not be broadcast to indexing result of shape (1,200)_**_

Previously, when working the model.plot() opened a new window showing the function(s) model, training data and the 95% confidence interval.

Sincerely,
Himakar Ganti

SebastianCallh:
Hello @himakarganti! I (and I assume several others) have experienced the same problem. Since you closed this issue I assume you figured it out? Would you mind sharing how you solved the plotting?

himakarganti:
Updating GPy to the latest version also updates Scipy to the latest version, which is having this issue on f2py; I wanted to point this because using an older version of GPy with the newer version of scipy fails as well.

SebastianCallh:
Thank you for your response. I spent yesterday trying to run with older versions of scipy and GPy, but to no avail. Do you have a working requirements.txt that you could share?

tyler-abbot:
I've re-opened this issue as it hasn't been solved. @himakarganti What is the f2py issue you mention? How did you fix your problem? Does GPy need to be fixed or simply wait for Scipy to be fixed?

darthdeus:
I'm running into this now as well after fiddling with versions. Does anyone know what is the actual cause, or at least the right way to fix this?

rwolst:
Someone more knowledgeable than me should check this, but the following fixes it for my application:

Change

    Xgrid[:,free_dims] = Xnew

to

    Xgrid[:,free_dims[0]] = Xnew.squeeze()

My reason for thinking this is probably ok, is that we are already inside an `if` block stating that `len(free_dims) == 1`. Furthermore, if `Xnew` cannot be squeezed to a 1D array then the original statement would have failed anyway.

I think the issue must be in a Numpy update that has changed the rules for assignment.

[...]

PR #766, opened 2019-06-18
Fix model.plot() to work with new Numpy update.

rwolst:
This fixes both the issues

https://github.com/SheffieldML/GPy/issues/723

and 

https://github.com/SheffieldML/GPy/issues/728

I am reasonably sure this doesn't break any backward compatibility with earlier Numpy versions. As per my previous comments in the first link:

> My reason for thinking this is probably ok, is that we are already inside an if block stating that len(free_dims) == 1. Furthermore, if Xnew cannot be squeezed to a 1D array then the original statement would have failed anyway.

> I think the issue must be in a Numpy update that has changed the rules for assignment.


ekalosak:
Looks like it does break backwards compatibility via CI test failure - I'm going to go ahead and close this due to inactivity.

PR #775, opened 2019-08-02
Corrected Multivariate Gaussian prior

monabf:
Corrected code of Multivariate Gaussian prior on hyperparameters, and added a little unit test for it. Seems to be working for me!


GPy/core/parameterization/priors.py
```diff
@@ -226,11 +233,11 @@ def pdf(self, x):
 
     def lnpdf(self, x):
         d = x - self.mu
-        return self.constant - 0.5 * np.sum(d * np.dot(d, self.inv), 1)
+        return self.constant - 0.5 * np.dot(d.T, np.dot(self.inv, d))
```
zhenwendai, line 241: This implementation does not work when x has more than one dimension.

monabf: Thanks for taking a look. I think you're right, but is the multivariate Gaussian prior not meant for vectors of hyperparameters anyway? If x is a vector it works, and I don't see why someone would want to put a multivariate Gaussian prior on a matrix of hyperparameters

monabf:
Hi everyone,
Could someone please take a look at this? I've been using it to put a multivariate Gaussian prior on vectors of parameters and it's been working well. If somebody needs an extension for more than a vector of parameters, then probably they should implement one, but multivariate Gaussians are only defined for random vectors so it will be another type of distribution. It would be really helpful for me if this pull request got into the main code! 
Thanks

monabf:
@zhenwendai I think I understood what you meant. Multivariate Gaussians are only defined for random vectors not matrices, which is what I was saying. But you want the implementation to work when the prior is set on x of shape (n, 1) and not only (n,) is that right? That should be fixed now, this implementation can be used for x of shape (n,1) or (n,) to set a multivariate Gaussian prior. I added a test for each case. This has been working well for me for a while and I would really need it to be merged into the main code, it would be great if you could take a look!
Thanks!

lawrennd:
@zhenwendai would you be happy with this to be merged now?

ekalosak:
@monabf can you rebase this on `SheffieldML:devel` HEAD? This looks good to merge and the CI has been updated - let's just get this to pass and it's good to go.

monabf:
Thanks @ekalosak, I merged the head of the devel branch into this pull request. Should be ready now, let me know if you need anything else!

ekalosak:
The merge from devel -> your branch is not showing up here. That's why the appveyor CI is failing: your branch has the python 2.7 build configured. The devel HEAD does not support python 2.7 nor does it require a passing build for python 2.7 - this is why I'd recommend rebasing on top of devel HEAD.

This is just a technical issue with the way the CI is configured in your outdated branch, not a problem with your contribution per se. Should be a simple fix 🤞

PR #783, opened 2019-09-05
Update self.num_data in GP when X is updated

MashaNaslidnyk:
self.num_data used to stay set to the old X. It should be updated when X is changed.

### DIFF
--- a/GPy/core/gp.py
+++ b/GPy/core/gp.py
@@ -43,8 +43,6 @@ class GP(Model):
             self.X = X.copy()
         else: self.X = ObsAr(X)
 
-        self.num_data, self.input_dim = self.X.shape
-
         assert Y.ndim == 2
         logger.info("initializing Y")
 
@@ -199,6 +197,14 @@ class GP(Model):
     def _predictive_variable(self):
         return self.X
 
+    @property
+    def num_data(self):
+        return self.X.shape[0]
+
+    @property
+    def input_dim(self):
+        return self.X.shape[1]
+
     def set_XY(self, X=None, Y=None):
         """
         Set the input / output data of the model
@@ -236,7 +242,6 @@ class GP(Model):
             else:
                 self.X = ObsAr(X)
 
-            self.num_data, self.input_dim = self.X.shape
         self.update_model(True)
 
     def set_X(self,X):
--- a/GPy/core/gp.py
+++ b/GPy/core/gp.py
@@ -235,6 +235,8 @@ class GP(Model):
                     self.link_parameter(self.X, index=index)
             else:
                 self.X = ObsAr(X)
+
+            self.num_data, self.input_dim = self.X.shape
         self.update_model(True)
 
     def set_X(self,X):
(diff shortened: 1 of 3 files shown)


zhenwendai (dismissed):

GPy/core/gp.py
```diff
@@ -235,6 +235,8 @@ def set_XY(self, X=None, Y=None):
                     self.link_parameter(self.X, index=index)
             else:
                 self.X = ObsAr(X)
+
+            self.num_data, self.input_dim = self.X.shape
```
zhenwendai: A better solution would be to change ```num_data``` and ```input_dim``` as property that return from the shape of ```X``` on-the-fly, because there may be other cases in future ```self.X``` getting changed.

MashaNaslidnyk: Not sure how that can be easily realised, with Python not having pointers... 
One (not best practice..) suggestion is to override `__setattr__` so `gp.X = X` calls `gp.set_XY(X)` under the hood:

```python
def __setattr__(self, name, value):
    if name == 'X':
        self.set_XY(X=value)
    elif name == 'Y':
        self.set_XY(Y=value)
    else:
        object.__setattr__(self, name, value)
```

API-changing alternatives: make `num_data` a function, or change `X` to `_X` (hinting at it being private and not to be directly fiddled with).

zhenwendai: What I mean is simply to do something like
```python
@property
def num_data(self):
    return self.X.shape[0]
```

In this way, ```self.num_data``` will be always up-to-date.

MashaNaslidnyk: Yes, but that's an API change. `gp.num_data` will no longer return the value, `gp.num_data() `will.  `gp.num_data` will simply refer to the method, so all code that uses it will break.

Toy example:

In [1]: import numpy as np

In [2]: class C():
   ...:     def __init__(self, a):
   ...:         self.a = a
   ...: 
   ...:     def shape(self):
   ...:         return self.a.shape
   ...:     

In [3]: c = C(np.array([2,2]))

In [4]: c.a.shape
Out[4]: (2,)

In [5]: c.shape()
Out[5]: (2,)

In [6]: c.shape
Out[6]: <bound method C.shape of <__main__.C object at 0x10ef11cf8>>

mmahsereci: Zhenwen's example won't break API. See [here](https://docs.python.org/3/library/functions.html#property). The property decorator turns the method into a getter. 

To see, you can modify your example by adding the property decorator to def shape(self) and it should work then.
```
class C():
    def __init__(self, a):
        self.a = a

    @property
    def shape(self)
        return self.a.shape
```

MashaNaslidnyk: oh, I didn't know about that decorator, and completely missed it in Zhenwen's example! Very cool. I'll modify the request, thanks Maren!

MashaNaslidnyk: Done. I had to make some changes in one class that inherits from GP, too. If anyone has a local class that inherits from GP and forcefully sets num_data / input_dim , this change will break it (they should be able to fix it easily, unless they rely on the mismatch between X.shape and num_data / input_dim for some reason). Let me know if this is a concern

Also, other Models have the same issue (e.g. TPRegression). I'm considering changing those too. Let me know of any thoughts on that.

PR #794, opened 2019-11-05
Stop using legacy bdist_wininst

hugovk:
The `bdist_wininst` file type is deprecated by PyPI:

* https://www.python.org/dev/peps/pep-0527/

It will probably be removed soon:

* https://github.com/pypa/warehouse/issues/6792


zhenwendai:
@mzwiessele Would this impact our deployment of wheels?

Issue #802, opened 2019-12-27
Rescale predictive_gradients when normalizer=True


ablancha:
Hello,

I was expecting `model.predictive_gradients` to return the gradients of `model.predict`, but it looks like it actually returns the gradients of `model._raw_predict`.  When `normalizer=True`, the user has to correct the results of `model.predictive_gradients` manually.  (See code below.)

This is problematic when `model.predictive_gradients` is used in an acquisition function in the context of Bayesian Optimization or Experimental Design, as in [GPyOpt](https://github.com/SheffieldML/GPyOpt) or [Emukit](https://github.com/amzn/emukit).

The easy solution is to edit the docstring of `model.predictive_gradients` to make clear that normalization is not included.  A more durable solution is to rescale the output of `model.predictive_gradients`, similar to `model.predict` or `model.predict_quantiles`.  (I suspect that `model.predict_jacobian` has the same problem but I haven't checked.)

If I am missing anything, please let me know.  

Thanks,
Antoine

```python
import numpy as np
import GPy

np.random.seed(2)

X = np.random.rand(15,3)
Y = np.random.rand(15,1)

normalize_Y = True
ker = GPy.kern.RBF(input_dim=X.shape[1])
model = GPy.models.GPRegression(X=X, Y=Y, kernel=ker,
                                normalizer=normalize_Y)

x = np.array([[-0.3, 0.1, 0.5]])
mu, var = model.predict(x)
mu_jac, var_jac = model.predictive_gradients(x)

# Finite-difference approximation
eps = 1e-8
mu_jac_num = np.zeros(x.shape[1])
var_jac_num = np.zeros(x.shape[1])
for ii in range(x.shape[1]):
    x_eps = x + eps*np.eye(1, x.shape[1], ii)
    mu_eps, var_eps = model.predict(x_eps)
    mu_jac_num[ii] = (mu_eps-mu)/eps
    var_jac_num[ii] = (var_eps-var)/eps

print('MU')
print(mu_jac.ravel())
print(mu_jac_num.ravel())
if normalize_Y:
    print(mu_jac.ravel() * model.normalizer.std)

print('VAR')
print(var_jac.ravel())
print(var_jac_num.ravel())
if normalize_Y:
    print(model.normalizer.inverse_variance(var_jac).ravel())
```

junhyeokahn:
I think you're right.
Any news or commits related to this?

ablancha:
[...]

PR #806, opened 2020-01-14
Fix predictive_gradients and posterior_covariance_between_points when normalizer=True

ablancha:
Addresses #802, with new tests added to `model_test.py`:

1. Add normalization in `model.predictive_gradients` so that it returns the gradients of `model.predict` even when `normalizer=True`.

2. Add normalization and likelihood in `model.posterior_covariance_between_points` so that it is consistent with the (co)variance from `model.predict` even when `normalizer=True`.

Fix 1 is important because `model.predictive_gradients` is used extensively in [GPyOpt](https://github.com/SheffieldML/GPyOpt) and [Emukit](https://github.com/amzn/emukit).  Fix 2 may be more controversial, although I suspect that most people expect `model.posterior_covariance_between_points(x,x)` to return the same result as `model.predict(x)[1]`.

### DIFF
--- a/GPy/core/gp.py
+++ b/GPy/core/gp.py
@@ -451,6 +451,15 @@ class GP(Model):
             alpha = -2.*np.dot(kern.K(Xnew, self._predictive_variable),
                                self.posterior.woodbury_inv)
             var_jac += kern.gradients_X(alpha, Xnew, self._predictive_variable)
+
+        if self.normalizer is not None:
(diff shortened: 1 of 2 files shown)


ekalosak:
Re: fix 2 we aught to have at least a minor version change if the semantics of this call are to change. But it _is_ a nice fix.

ablancha:
Responded in #802.

PR #817, opened 2020-03-09
fix ImportError in likelihood.py

kae-mihara:
in function log_predictive_density_sampling


zhenwendai:
Could you please bump the required version of scipy to the minimal version that supports the new api of logsumexp in ```setup.py```?

PR #829, opened 2020-05-03
Use super().__init__ consistently

jbect:
Use `super().__init__` consistently, instead of sometimes calling base class `__init__` directly.

The syntax with `super` was already used almost everywhere in GPy (162 occurences before the proposed patch, 200 occurences after).


lawrennd (approved):
Thanks ... that really tidies things up!

PR #842, opened 2020-06-17
Fix scipy depency problem

jbect:
Travis builds currently fail with python 3.5 because scipy 1.5 is used (instead of 1.3.x or 1.4.x).

This is a new problem caused by the release of scipy 1.5.0rc1 in May.


lawrennd:
Thanks for fixing this.

Issue #844, opened 2020-06-18
matplotlib.cbook.mkdirs has been removed from matplotlib 3.2


jbect:
`matplotlib.cbook.mkdirs`, which is used in [plotting_tests.py](https://github.com/SheffieldML/GPy/blob/devel/GPy/testing/plotting_tests.py#L90), was deprecated in matplotlib 3.1 and has been removed from 3.2.

This causes errors in some CI jobs:
![image](https://user-images.githubusercontent.com/17221892/85017713-3dd25400-b16c-11ea-8eb1-a1df14376d47.png)

PR #845, opened 2020-06-18
Use os.makedirs instead of matplotlib.cbook.mkdirs

jbect:
Use `os.makedirs` instead of `matplotlib.cbook.mkdirs`.  This PR fixes issue #844


lawrennd (approved):
Looks good, thanks for fixing!

PR #863, opened 2020-09-24
Linting and static analysis (pylint, codacy)

bobturneruk:
`pylint` and other static analysis tools can provide helpful suggestions to improve `GPy`. Here, I propose a `pylint` config file that limits some checks where they deviate from `GPy` policy and propose using the Codacy platform as a simple means of gaining insight into what changes might be useful:

https://app.codacy.com/gh/RSE-Sheffield/GPy/dashboard

If this were merged into `devel`, an admin would need to do the codacy setup as permissions are needed - this example is on our fork.

Further customisation of codacy checks in line with `GPy` policy and community requirements would be essential.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,590 @@
+[MASTER]
+
+# A comma-separated list of package or module names from where C extensions may
+# be loaded. Extensions are loading into the active Python interpreter and may
+# run arbitrary code.
+extension-pkg-whitelist=
+
+# Specify a score threshold to be exceeded before program exits with error.
+fail-under=10.0
+
+# Add files or directories to the blacklist. They should be base names, not
+# paths.
+ignore=CVS
+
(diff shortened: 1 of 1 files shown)


ekalosak:
I duplicated this (in spirit, at least) in https://github.com/SheffieldML/GPy/issues/910.

The pylint solution looks good, I'm not sure why we'd use that rather than flake8 though.

So, @bobturneruk , are you proposing adding a link to Codacity from the README.md, adding linting to CI as a blocking check, or something else? Keeping the code quality tools up-to-date is important, but I'm concerned we're snowballing everyone's favorite tool in the CI as-is.

bobturneruk:
Hi @ekalosak! I think any static analysis would be a good plan. I was proposing adding this in CI as a guide to contributors, with Codacy as the platform. The more prescriptive approach of mandating black is fine if it works for this community.

PR #864, opened 2020-09-24
Fix CI

bobturneruk:
There was a problem running `travis_tests.py` on my computer and in appveyor whereby a parameter of `matplotlib` that it appears is not present in more recent versions was set. This PR resolves this by changing the call to matplotlib. An alternative solution would be to more tightly define the environments used during testing. 

Anyway, long term I think it might be better to try and support more recent versions of dependencies.


lawrennd (approved):
Currently tests fail without this fix. Integrating

PR #867, opened 2020-09-28
Improve kernel tests

bobturneruk:
Kernel tests are executed against each kernel using the [`check_kernel_gradient_functions`](https://github.com/SheffieldML/GPy/blob/ec20f9ed3a6d98e278a35114e592f1ea098b5531/GPy/testing/kernel_tests.py#L147) function which, in itself, calls and executes a number of tests. Currently, if one of these tests fails or errors, **the subsequent tests do not run**. This PR modifies the code such that if any one test fails or errors, a failure is reported, but the additional tests still run. This means that a developer can fix problems causing tests to fail in any order they like.

### DIFF
--- a/GPy/testing/kernel_tests.py
+++ b/GPy/testing/kernel_tests.py
@@ -176,8 +176,7 @@ def check_kernel_gradient_functions(kern, X=None, X2=None, output_ind=None, verb
     if not result:
         print(("Positive definite check failed for " + kern.name + " covariance function."))
         pass_checks = False
-        #assert(result)
-        #return False
+
 
     if verbose:
         print("Checking gradients of K(X, X) wrt theta.")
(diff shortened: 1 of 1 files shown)


drj11 (commented):
sure, it seems reasonable for now.

Really it seems to me that these tests should be separated so that the test framework can "see" them individually and run and report separately.

Possibly using _parameterized tests_: https://docs.pytest.org/en/stable/parametrize.html#basic-pytest-generate-tests-example

Tests that can't succeed should be _skipped_: https://docs.pytest.org/en/stable/skipping.html

I'm not suggesting you do that in this PR, it's another PR. But note that doing either of things i suggest above involves really pinning down which test framework you want to go forward with.

bobturneruk:
Thanks @drj11 - I completely agree, it's (hopefully) an improvement, but not ideal. If the resources are available, I'd favour moving from `nosetests` to `pytest` as the latter has more current support and is likely to have more longevity.

PR #868, opened 2020-09-30
Update README.md

Atharvajadhao:


bobturneruk:
The CI checks will fail for all PRs at present, see https://github.com/SheffieldML/GPy/pull/864

This could do with attention from an admin or maintainer, when time allows.

lawrennd (approved):
Minor edits only. Thanks for providing! These are approved.

PR #872, opened 2020-11-16
Fix description of exponentiated quadratic covariance function.

lawrennd:


GPy/kern/src/stationary.py
```diff
@@ -641,6 +641,32 @@ def dK_dr(self, r):
...
+class Cosine(Stationary):
+    """
+    Cosine Covariance function
+    
+    .. math::
+
+        k(r) = \sigma^2 \cos(r)
+
+    """
     def __init__(self, input_dim, variance=1., lengthscale=None, ARD=False, active_dims=None, name='Cosine'):
```
ekalosak: You've defined the `class Cosine(Stationary)` twice.

ekalosak (changes requested):
@lawrennd looks like you defined the Cosine kernel twice - aside from that this LGTM.

GPy/kern/src/stationary.py
lawrennd: Doh! I'll fix.

lawrennd: Fixed!

PR #903, opened 2021-04-15
Fix GPy.priors.InverseGamma

ekalosak:
fixed InverseGamma prior: beforehand, it was a child class of Gamma but it defined a broken __new__ method of its own. Now, it just inherits Gamma's __new__; also added a test that ensures the InverseGamma can be instantiated and integrated into a GPy model

### DIFF
--- a/GPy/core/parameterization/priors.py
+++ b/GPy/core/parameterization/priors.py
@@ -358,13 +358,15 @@ class InverseGamma(Gamma):
     domain = _POSITIVE
     _instances = []
 
-    def __init__(self, a, b):
-        self._a = float(a)
-        self._b = float(b)
-        self.constant = -gammaln(self.a) + a * np.log(b)
-
     def __str__(self):
         return "iGa({:.2g}, {:.2g})".format(self.a, self.b)
+
+    def summary(self):
+        return {}
+
+    @staticmethod
+    def from_EV(E, V):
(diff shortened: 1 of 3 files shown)


ekalosak:
Also, I'm curious why GPy.core.Model is Priorizable and has a set_prior() method - it's not a parameter itself.

GPy/core/parameterization/priors.py
```diff
@@ -357,24 +357,17 @@ class InverseGamma(Gamma):
     """
     domain = _POSITIVE
     _instances = []
-    def __new__(cls, a=1, b=.5): # Singleton:
```
lawrennd, line 360: Does removing this mean that those default settings are lost? Or are they the same as the original Gamma?

lawrennd (approved):
Looks fine to me. 

I haven't dug into the Gamma prior though and checked that it interprets a and b in the same way when constructing.

GPy/core/parameterization/priors.py
ekalosak: They're identical to the defaults from `class Gamma(Prior)`'s `__new__` as defined on `GPy/core/parameterization/priors.py#L291` in `devel:HEAD` at time of writing (i.e. at commit `ff82f12c3d321bfc3ce6615447fad25aea9de6bd`).

Issue #896, opened 2021-03-08
is operator with literal warning 3.8


charelstoncrabb:
Running version `Python 3.8.7` and `GPy==1.9.9`:

```
 SyntaxWarning: "is" with a literal. Did you mean "=="?
  if symmetry_type is 'odd':
/home/$USER/.pyenv/versions/3.8-dev/envs/venv/lib/python3.8/site-packages/GPy/kern/src/symmetric.py:58: SyntaxWarning: "is" with a literal. Did you mean "=="?
  elif symmetry_type is 'even':
 /home/$USER/.pyenv/versions/3.8-dev/envs/venv/lib/python3.8/site-packages/GPy/models/mrd.py:309: SyntaxWarning:"is not" with a literal. Did you mean "!="?
```

ekalosak:
Good catch, this is still present in 1.10.0 too. If you cut a PR for this it'd get merged quick.

PR #911, opened 2021-05-17
addressed @charelstoncrabb issue reported in #896

ekalosak:
Addresses #896


lawrennd (approved):
Changes look good to me.

ekalosak:
Thanks for the review lawrennd.

This PR has an `f`-string in it, though - that makes Python3.5 unhappy. I'm going to wait for a day or two until we've had sufficient time for objection in the "should we drop 3.5 support?" issue thread before dropping 3.5 and merging this PR.

lawrennd:
Makes sense!

PR #916, opened 2021-05-19
Fixing errors causing GPLVM cmu_mocap example to not run properly.

lawrennd:
Various errors in the dimensionality reduction example cmu_mocap that required fixes elsewhere including:

1) Use of == to compare objects to None.
2) Missing ability to normalise inputs to GPLVM (as available in base class GP)
3) Standardize() normaliser class not checking for when standard deviations zero, leading to a silent "nan" error (fixed by warning of standard deviation zero, and setting std to 1.0
4) Changes to output of plot_latent() which meant it wasn't returning an axes object as the demo expected (fix in demo code).

### DIFF
--- a/GPy/util/normalizer.py
+++ b/GPy/util/normalizer.py
@@ -4,7 +4,7 @@ Created on Aug 27, 2014
 @author: Max Zwiessele
 '''
 import numpy as np
-
+import warnings
 
 class _Norm(object):
     def __init__(self):
--- a/GPy/util/normalizer.py
+++ b/GPy/util/normalizer.py
@@ -90,7 +90,7 @@ class Standardize(_Norm):
         Y = np.ma.masked_invalid(Y, copy=False)
         self.mean = Y.mean(0).view(np.ndarray)
         self.std = Y.std(0).view(np.ndarray)
-        if np.any(self.std) == 0:
+        if np.any(self.std == 0):
             self.std[np.where(Y_std==0)]=1.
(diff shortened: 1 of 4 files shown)


lawrennd:
@ekalosak if you could do a sanity check on this that would be great!

ekalosak (approved):
LGTM.

I wonder why you're replacing `self.std[np.where(self.std==0)]` with `1.` rather than `0.0000001` - might be worth an inline comment, but not a blocking suggestion.

lawrennd:
Yes, good question. Could also replace it with EPS or something similar. 

I think it's somewhat arbitrary and can't see strong arguments for one or the other ... I can add a comment to that effect.

GPy/util/normalizer.py
```diff
@@ -90,6 +90,9 @@ def scale_by(self, Y):
         Y = np.ma.masked_invalid(Y, copy=False)
         self.mean = Y.mean(0).view(np.ndarray)
         self.std = Y.std(0).view(np.ndarray)
+        if np.any(self.std == 0):
+            warnings.warn("Some values of Y have standard deviation of zero. Resetting to 1.0 to avoid divide by zero errors.")
+            self.std[np.where(self.std==0)]=1.
```
lawrennd:
```suggestion
           # Choice of setting to 1.0 is somewhat arbitrary. It avoids a divide by zero error, but setting to EPS would also do this. Don't have strong reasons for choosing 1.0, it was just first instinct 
            self.std[np.where(self.std==0)]=1. 
```

lawrennd:
Merging commit after approval checks from @ekalosak

PR #917, opened 2021-05-20
linted GPy/util using flake8 and formatted using black

ekalosak:
No change in functionality, just showing what the result of black & flake8 is on this codebase in a relatively untouched (recently) corner of the code. Linting and formatting is an incremental process, and this PR gets that effort off the ground.

### DIFF
--- a/GPy/util/block_matrices.py
+++ b/GPy/util/block_matrices.py
@@ -2,36 +2,39 @@
 # Licensed under the BSD 3-clause license (see LICENSE.txt)
 import numpy as np
 
+
 def get_blocks_3d(A, blocksizes, pagesizes=None):
     """
     Given a 3d matrix, make a block matrix, where the first and second dimensions are blocked according
     to blocksizes, and the pages are blocked using pagesizes
     """
-    assert (A.shape[0]==A.shape[1]) and len(A.shape)==3, "can't blockify this non-square matrix, may need to use 2d version"
+    assert (A.shape[0] == A.shape[1]) and len(
+        A.shape
+    ) == 3, "can't blockify this non-square matrix, may need to use 2d version"
     N = np.sum(blocksizes)
     assert A.shape[0] == N, "bad blocksizes"
     num_blocks = len(blocksizes)
-    if pagesizes == None:
-        #Assume each page of A should be its own dimension
-        pagesizes = range(A.shape[2])#[0]*A.shape[2]
+    if pagesizes is None:
+        # Assume each page of A should be its own dimension
+        pagesizes = range(A.shape[2])  # [0]*A.shape[2]
     num_pages = len(pagesizes)
     B = np.empty(shape=(num_blocks, num_blocks, num_pages), dtype=np.object)
-    count_k = 0
-    #for Bk, k in enumerate(pagesizes):
     for Bk in pagesizes:
         count_i = 0
         for Bi, i in enumerate(blocksizes):
             count_j = 0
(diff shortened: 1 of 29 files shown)


GPy/util/block_matrices.py
```diff
@@ -2,36 +2,39 @@
...
 import numpy as np
 
+
 def get_blocks_3d(A, blocksizes, pagesizes=None):
     """
     Given a 3d matrix, make a block matrix, where the first and second dimensions are blocked according
     to blocksizes, and the pages are blocked using pagesizes
     """
-    assert (A.shape[0]==A.shape[1]) and len(A.shape)==3, "can't blockify this non-square matrix, may need to use 2d version"
+    assert (A.shape[0] == A.shape[1]) and len(
```
lawrennd, line 11: This one's a bit ugly ... but it also makes me wonder if assert should be used in these cases. 

Would the right thing be to raise a ValueError?

GPy/util/choleskies.py
```diff
@@ -73,34 +80,42 @@ def _backprop_gradient_pure(dL, L):
...
+    triangulars = (n ** 2 + n) / 2
     keep = []
     for i in range(rank):
         keep.append(triangulars[i:] + i)
-    #add the diagonal
-    keep.append(triangulars[1:]-1)
-    keep.append((size**2 + size)/2 -1)# the very last element
+    # add the diagonal
+    keep.append(triangulars[1:] - 1)
+    keep.append((size ** 2 + size) / 2 - 1)  # the very last element
```
lawrennd, line 115: That's funny ... spaces around the power operator give me the shivers a bit! But Agree spaces around the operator better in general and better to be consistent!

lawrennd (approved):
I like the idea of doing this. 

I haven't looked through every file, but in general it looks good. So I'm somewhat relying on black doing the right thing to approve.

I also think we need a big sort out of these files, some of this functionality can/should be moved to other packages. datasets.py is already in the pods toolbox, I'm tidying that up for a new release as we speak.

GPy/util/block_matrices.py
ekalosak: Yeah, the right thing would be `ValueError`. In general, I find the following block extracted from https://wiki.python.org/moin/UsingAssertionsEffectively to be a good guide.

"""
Assertions should *not* be used to test for failure cases that can occur because of bad user input or operating system/environment failures, such as a file not being found. Instead, you should raise an exception, or print an error message, or whatever is appropriate. One important reason why assertions should only be used for self-tests of the program is that assertions can be disabled at compile time.
"""

Following the notion that `assert` shouldn't address bad user input - in this case, the shape of `A` - it should indeed instead be a `ValueError`.

That being said, this suggestion is outside the scope of this PR - I'm just making linting and styling changes here. I'll file an issue, one sec...

ekalosak: https://github.com/SheffieldML/GPy/issues/922

GPy/util/choleskies.py
ekalosak: I like `black` because it has style opinions so I don't have to 😅 It's not a panacea, but it _is_ identical everywhere. If this stylistic imposition is cool, please resolve; otherwise we should discuss.

ekalosak:
Ah, I messed up a bunch of the imports that are intended to be re-imported elsewhere. I'll fix that up and add the `  # noqa: F401` suffix to those lines that will indicate they're supposed to be there even if they're not used in the importing module.

PR #918, opened 2021-05-21
Fix code that was relying on Python 2 integer division

lawrennd:


lawrennd:
@ekalosak if you have time to give this a quick look. There's only one commit (I think I messed up the last by rebasing ...) 

The commit is fixing a reliance on the python2 integer division for a reshape. Add code to check reshape is valid and throw an error message explaining what's wrong.

ekalosak (approved):
LGTM, none of the following suggestions would block a merge:
- I'd suggest adding a test on the `process_values()` function to ensure we don't get blindsided by changes to numpy or the language down the line.
- Also, it might be a bit more informative to raise a new exception from your catch block e.g.
```
except ValueError as e:
  raise ValueError("Passed values to stick_show need ...") from e
```

As long as this is "Squash and merge"d, the commit history isn't a problem 👍

lawrennd:
Cool, thanks Eric. I didn't know about 'raise from', that's pretty neat.

I've also run black (maybe I should have done that separately). 

I'll merge now.

PR #928, opened 2021-05-28
Make examples files check if matplotlib is available.

lawrennd:


ekalosak (approved):
LGTM :shipit: , might be useful to only calculate `MPL_AVAILABLE` once in e.g. the plotting module so the code isn't repeated, but that's not a blocker.

lawrennd:
Yes, that would be better. 

I'll merge for now and maybe there's a rethink of plotting needed in general ... something for the new discussions channel perhaps?

Issue #925, opened 2021-05-28
error in prior.plot()


flydream0428:
When the prior.plot() is called, it gives the error below, 
```
gamma_prior = GPy.priors.Gamma.from_EV(1.5, 0.7)
gamma_prior.plot()
```
> AttributeError: 'Rectangle' object has no property 'normed'

This should be fixed by replacing "normed" with "density".

ekalosak:
Thanks for the productive post @flydream0428 💪 Would you mind cutting a PR that fixes this?

flydream0428:
@ekalosak Sure, I will have a go

ekalosak:
Done, thank you.

PR #929, opened 2021-05-30
Update priors_plots.py

flydream0428:
Change pb.hist(rvs, 100, normed=True) to pb.hist(rvs, 100, density=True) in the file priors_plots.py.
This is a fix to https://github.com/SheffieldML/GPy/issues/925


ekalosak:
There wasn't a test for this code beforehand, so I'm not going to drag the contribution by expanding the scope to require a new test.

PR #938, opened 2021-07-16
Fix ignoring arguments passed to save_model.

peterscherpelz:
Before this change, defaults were passed to _save_model, rather than the
correct arguments the user supplied to save_model.


ekalosak:
This looks good - to merge, we need a test that covers this code. That'll keep us from making the same mistake in the future. Please let me know if you need any help writing the test. I don't think it will be too onerous :)

PR #951, opened 2021-10-29
to_dict() and from_dict() functionality for Coregionalize Kernel and MixedNoise Likelihood class

gehbiszumeis:
These are mandatory prerequites for tracking `GPy.models.GPCoregionalizedRegression` models with Machine learning lifecycle tools (e.g mlflow)

### DIFF
--- a/GPy/core/gp.py
+++ b/GPy/core/gp.py
@@ -134,9 +134,10 @@ class GP(Model):
         if self.mean_function is not None:
             input_dict["mean_function"] = self.mean_function.to_dict()
         input_dict["inference_method"] = self.inference_method.to_dict()
-        #FIXME: Assumes the Y_metadata is serializable. We should create a Metadata class
+        # TODO: We should create a Metadata class
         if self.Y_metadata is not None:
-            input_dict["Y_metadata"] = self.Y_metadata
+            # make Y_metadata serializable
+            input_dict["Y_metadata"] = {k: self.Y_metadata[k].tolist() for k in self.Y_metadata.keys()}
         if self.normalizer is not None:
             input_dict["normalizer"] = self.normalizer.to_dict()
         return input_dict
--- a/GPy/core/gp.py
+++ b/GPy/core/gp.py
@@ -163,9 +163,10 @@ class GP(Model):
             input_dict["mean_function"] = mean_function
         input_dict["inference_method"] = GPy.inference.latent_function_inference.LatentFunctionInference.from_dict(input_dict["inference_method"])
 
-        #FIXME: Assumes the Y_metadata is serializable. We should create a Metadata class
-        Y_metadata = input_dict.get("Y_metadata")
-        input_dict["Y_metadata"] = Y_metadata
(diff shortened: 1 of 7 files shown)


ekalosak:
Looks like a CI error, otherwise LGTM. Any thoughts on the CI error?
Also, where're the `from_dict` implementations? Or, are they inherited?

gehbiszumeis:
> Looks like a CI error, otherwise LGTM. Any thoughts on the CI error? 

Yes, according to AppVeyor it seems, that it fails already doing the build for almost all environments. It cannot install a package apparently

`ERROR conda.core.link:_execute(502): An error occurred while installing package 'defaults::charset-normalizer-2.0.4-pyhd3eb1b0_0'.`

No idea why, I didn't touch the `requierements.txt`. Only for the py3.5 environment it runs past this step and breaks when running the code coverage test

`GPy: Saving user configuration file to C:\Users\appveyor\.config\GPy\user.cfg
coverage run travis_tests.py
........S............S...............................................................................................................................................................................................................................................................................................................................S.............................................................................................Command exited with code -1073741819`

> Also, where're the `from_dict` implementations? Or, are they inherited?

Yes, the corresponding `from_dict()` methods of the super classes are used. I added only the `_build_from_input_dict` methods to add some specific parameters to the corresponding `input_dict`.

ekalosak:
Fun enough, the errors are different across Python versions.
## Python 3.5
Screen Shot 2021-11-10 at 1 14 52 PM
## Python 3.6 thru 3.9
Screen Shot 2021-11-10 at 1 16 56 PM
## Things to try
- The `conda clean` command in the 3.6 error message
- Looks like the 3.5 error code shows up [here](https://github.com/mapbox/node-sqlite3/issues/1318) as well, something to think about...

gehbiszumeis:
> Things to try - The `conda clean` command in the 3.6 error message - Looks like the 3.5 error code shows up [here](https://github.com/[mapbox/node-sqlite3/issues/1318](https://github.com/mapbox/node-sqlite3/issues/1318)) as well, something to think about...

The 3.5 error disappeared, the >3.6 errors persist. I tried quite some stuff manipulating the way conda is trying to install packages, with no success yet. After all the trying it seems, that the `charset-normalizer` package has only a `noarch` version and no `win64` version. 

What really puzzles me is that the build fails only for this PR and not for others, so it cannot be some general bug in `charset-normalizer` or similar. It seems that for my build it tries to install that package, while for other PR builds it doesn't. That's weird

ppk42:
I brought the appveyor build back to life. The pull request is still open. But, you'll find the changes in the last commit. (https://github.com/SheffieldML/GPy/pull/956/commits/a93c4ace5f884db4f55e0faffd9f3287d5e86f35)

ekalosak:
Brilliant work both, thank you for the contributions!
Because the code coverage is still not really a priority, passing builds and tests is totally sufficient so I'm going to merge this.
@lawrennd @apaleyes <- please give this an 👁️ so we can remove it if it's bad.

lawrennd:
Thanks @ekalosak and @ppk42!

Issue #955, opened 2021-11-27
Fix AppVeyor (Windows) CI Build


ppk42:
I was aware of pull request https://github.com/SheffieldML/GPy/pull/951. One of the prerequisites that must be fulfilled is a AppVeyor buld for the Windows version. The build only works for Python 3.5 but not for Python versions 3.6 up to 3.9. 

The build is essential for further development. If it will not work, no pull requests can/should be accepted. 

With the known facts, it seems that the build did not fail because of the code changes. There is evidence for the assumption that the build failed because of incompatibilities of the miniconda version and the `conda update conda` command. 

It is remarkable that the working case and the not working cases use different installation folders of the miniconda environment and, all not working cases use the same miniconda installation.  

I would like to check these thoughts by forking the repo and adding pull requests with changes only for the appveyor.yml file that hopefully bring appveyors build to life again. ;-). It could also be possible that the ApVeyor configuration must be changed (e.g. another Visual Studio VM), but that's beyond my means. Only the owner of the AppVeyor project is granted to change it.

ppk42:
The fix is documented in the PR https://github.com/SheffieldML/GPy/pull/956 and by the commits coments. TLDR; The outdated miniconda version 36 has issues with the `conda update -q conda` command. Fix: Use the C:\Miniconda3-x64\ folder then, conda update works. But, for python 3.7 and above a higher matplotlib version is used. Because of changes in the matplotlib the build failed for Python 3.7 but not for 3.8, 3.9 (probably caused by changes in the parser/compiler PEP570??). I decided to freeze the matplotlib version to 3.0.0 for python 3.5 and 3.3.4 for python 3.6, 3.7, 3.8 and 3.9. From what I saw in the code there is a problem in GPy with the wrong matplotlib version that is not covered by tests. That should be further investigated.

ekalosak:
Thanks @ppk42 .

PR #956, opened 2021-11-27
bring the appveyor build to life again

ppk42:
I suspect that an appveyor build is started by a webhook on Github (or by the pull request). 
I see changing the appveyor.yml with a subsequent pull request as the only way to isolate the error and to try out ideas.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -25,7 +25,8 @@ install:
  - conda config --set always_yes yes --set changeps1 no
  - conda update -q conda
  - conda info -a
- - "conda create -q -n build-environment python=%PYTHON_VERSION% numpy scipy matplotlib"
+# github issue #955: freez build versions of numpy, scipy and matplotlib
+ - "conda create -q -n build-environment python=%PYTHON_VERSION% numpy=1.19.2 scipy=1.5.2 matplotlib=3.3.4"
  - activate build-environment
  # We need wheel installed to build wheels
  - python -m pip install wheel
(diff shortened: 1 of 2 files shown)


ppk42:
I found a miniconda environment that does not fail for `conda update`. But now there are errors with the Python 3.7 build. 

I inspected the code: Tests use a private method from the _AxesBase class in the matplot lib. This method hat two keyword params `xdata` and `ydata`. The signature of the method changes from version 3.3.4 (used for the Python 3.6 build) to version 3.4.3 for python 3.7, 3.8 and 3.9 builds. 

In matplot 3.4.3 the keyword arguments have been removed. Python 3.7 raises an error because of the missing arguments. Higher Python version do not report an error, even though the wrong method signature was used too. This is dagenrous and until we find a better solution I sugesst to freez the matplotlib version. 

I'll try to freeze the matplolib version and also the scipy and numpy version. For a check I need to perform the PR again. I'll mark all changes in the setup.py and appveyor.yml with the corresponding issue number. https://github.com/SheffieldML/GPy/issues/955

ppk42:
Frezing the numpy and scipy version was a bad idea. The 3.5 build crashes on that. So I decided to use a conditional aproach  for matplotlib only (differnt for 3.5 at one side and all above 3.6 at the other).

ppk42:
Gehbiszumeis has adopted my changes in his pull request (https://github.com/SheffieldML/GPy/pull/951). The explicit merging of this pull request is not necessary. For documentation reasons, I will not delete the PR, but it can be closed.

ppk42:
Of cause, if for any reason #951 will not be accepted then, the appveyor build will stay broken.

Issue #953, opened 2021-11-12
TypeError: _process_unit_info() got an unexpected keyword argument 'xdata'


ozturkc:
I have tried to install GPy via "Troubleshooting installation problems" guide but I had "TypeError: _process_unit_info() got an unexpected keyword argument 'xdata'" error in MacOS terminal. How can I fix this? 
Thank you.

ekalosak:
## Please post the full traceback log, your GPy version, and your python version
We're here to help but can't do much without these key pieces.

## Looking briefly
<img width="1020" alt="Screen Shot 2021-11-22 at 12 39 56 PM" src="https://user-images.githubusercontent.com/6068321/142909334-6ddea9ea-9e9b-443d-bceb-3cab95c96fdd.png">
It's all matlotlib... So... Do you have `matplotlib` installed? Just a hunch while we await more detail from your end.

PR #960, opened 2021-12-14
Fix for matplotlib>=3.4.0

timovwb:
The arguments to `_process_unit_info` changed in matplotlib 3.4.0 (current latest is 3.5.1). This pull request makes the required change.

Note that this will of course break for installations with matplotlib<3.4.0.

Related issue: #953

### DIFF
--- a/GPy/plotting/matplot_dep/plot_definitions.py
+++ b/GPy/plotting/matplot_dep/plot_definitions.py
@@ -241,8 +241,8 @@ class MatplotlibPlots(AbstractPlottingLibrary):
             except ImportError:
                 from matplotlib.mlab import contiguous_regions
             # Handle united data, such as dates
-            ax._process_unit_info(xdata=X, ydata=y1)
-            ax._process_unit_info(ydata=y2)
+            ax._process_unit_info([("x", X), ("y", y1)], convert=False)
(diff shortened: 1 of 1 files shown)


ekalosak:
Make sure the dependency specifications reflect the deprecation of matplotlib<3.4.0 and then this looks good.

i-aki-y:
@timovwb Can I take over this fix?

@ekalosak 
The current setup.conf seems to specify matplotlib version 3.0.0 for python 3.5. 
Does this mean we need to stop support for py3.5 to apply this PR?

https://github.com/SheffieldML/GPy/blob/f63ed48b0d4bf043868a0a1729d28003c13fe263/setup.py#L120-L126

Or do we need other workarounds? ex.
```py
try:
    # for matplotlib <= 3.3.4
    ax._process_unit_info(xdata=X, ydata=y1)
    ax._process_unit_info(ydata=y2)
except TypeError:
    # for matplotlib > 3.4.0
    ax._process_unit_info([("x", X), ("y", y1)], convert=False)
    ax._process_unit_info([("y", y2)], convert=False)
```

timovwb:
@i-aki-y Definitely, go ahead.

PR #1000, opened 2023-01-11
pep508 dependency specification

jccalvojackson:
the current setup.py does not conform to pep508. Hence it can lead to issues like [this](https://github.com/python-poetry/poetry/issues/7340).

This PR rewrites the dependencies using pep508

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -126,14 +126,10 @@ install_requirements = [
     "cython>=0.29",
     'scipy>=1.3.0,<1.5.0;python_version<"3.6"',
     'scipy>=1.3.0;python_version>="3.6"',
-]
-
-matplotlib_version = [
     'matplotlib==3.0.0;python_version<"3.6"',
     'matplotlib==3.3.4;python_version>="3.6"',
 ]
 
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -117,13 +119,22 @@ def ismac():
...
+    'scipy>=1.3.0,<1.5.0;python_version<"3.6"',
+    'scipy>=1.3.0;python_version>="3.6"',
+]
+
+matplotlib_version = [
+    'matplotlib==3.0.0;python_version<"3.6"',
+    'matplotlib==3.3.4;python_version>="3.6"',
+]
+
+install_requirements += matplotlib_version
```
dimbleby: this makes matplotlib an unconditional dependency, which I don't think it was previously

jccalvojackson: you are right.

ekalosak:
This PR is a bit old, due to lapses on the fault of maintainers for reviewing PRs and merging them. That said, There's a conflict with the main branch that must be resolved. WIthout that resolution soon (next week or so) I'd recommend closing this PR without merging.

PR #1003, opened 2023-01-29
Add Wiener Velocity kernel

Duncan10:
This PR contributes an implementation for the 1D Wiener velocity kernel. The Wiener velocity corresponds to a once integrated Wiener process (Brownian motion) as described by Solin in "Stochastic Differential Equation Methods for Spatio-Temporal Gaussian Process Regression", 2016. URL: http://urn.fi/URN:ISBN:978-952-60-6711-7


ekalosak:
LGTM @zhenwendai @javiergonzalezh et al. what do you think?

PR #1011, opened 2023-03-15
Fix issues encountered in modern python versions

MartinBubel:
The constraints on the `scipy` version are quite restrictive and seem to cause linker errors on blas/lapack when combined with modern `numpy` versions.

Also, datatype `np.bool` does not exist in modern `numpy` versions, while `bool` exists for all python versions.

### DIFF
--- a/GPy/core/__init__.py
+++ b/GPy/core/__init__.py
@@ -77,7 +77,7 @@ def randomize(self, rand_gen=None, *args, **kwargs):
     # now draw from prior where possible
     x = self.param_array.copy()
     [np.put(x, ind, p.rvs(ind.size)) for p, ind in self.priors.items() if not p is None]
-    unfixlist = np.ones((self.size,),dtype=np.bool)
+    unfixlist = np.ones((self.size,),dtype=bool)
     from paramz.transformations import __fixed__
     unfixlist[self.constraints[__fixed__]] = False
     self.param_array.flat[unfixlist] = x.view(np.ndarray).ravel()[unfixlist]
--- a/GPy/inference/mcmc/samplers.py
+++ b/GPy/inference/mcmc/samplers.py
@@ -38,7 +38,7 @@ class Metropolis_Hastings(object):
     def sample(self, Ntotal=10000, Nburn=1000, Nthin=10, tune=True, tune_throughout=False, tune_interval=400):
         current = self.model.optimizer_array
         fcurrent = self.model.log_likelihood() + self.model.log_prior() 
-        accepted = np.zeros(Ntotal,dtype=np.bool)
+        accepted = np.zeros(Ntotal,dtype=bool)
         for it in range(Ntotal):
             print("sample %d of %d\r"%(it+1,Ntotal),end="")
             sys.stdout.flush()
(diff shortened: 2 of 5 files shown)


apaleyes:
Thanks for picking this up, just run into similar issue with numpy 1.24!

There are a few more places where `np.bool` is used, can you please update those too in this PR? https://github.com/SheffieldML/GPy/search?q=np.bool

MartinBubel:
> Thanks for picking this up, just run into similar issue with numpy 1.24!
> 
> There are a few more places where `np.bool` is used, can you please update those too in this PR? https://github.com/SheffieldML/GPy/search?q=np.bool

Thanks for checking. I just replaced the remainder usages of `np.bool`. Should be all good right now.

However, the changes I made to the `scipy` version require python >= 3.6.
The question now is whether you want to keep downward-compatibility below python 3.6 (if possible at all), considering that python 3.5 has reached end of life by quite some time now.
```
DEPRECATION: Python 3.5 reached the end of its life on September 13th, 2020. Please upgrade your Python as Python 3.5 is no longer maintained. pip 21.0 will drop support for Python 3.5 in January 2021. pip 21.0 will remove support for this functionality.
```

apaleyes:
Thank you for addressing it so quickly, @MartinBubel ! You ask a good question, which is not for me to answer - I do not maintain GPy itself, merely a package that depends on it. Personally, I see nothing wrong with dropping support for 3.5

zhenwendai:
I agree that we should drop the support for Python 3.5. Thanks @MartinBubel for the PR.

apaleyes:
Thanks for taking care of this, @zhenwendai !

apaleyes:
Unfortunately, I still have to nag about this a bit more. This page shows all available options to install GPy: https://pypi.org/simple/gpy/

The 1.11.0 version has only one very specific wheel, meaning it won't install on most platforms. Say, it doesn't install on my Ubuntu:

> Skipping link: none of the wheel's tags (cp35-cp35m-win_amd64) are compatible (run pip debug --verbose to show compatible tags): https://files.pythonhosted.org/packages/89/66/532048790189b5b63c17f5f11532dd4b7aff5a2c4ae9ea02299b4df55cc4/GPy-1.11.0-cp35-cp35m-win_amd64.whl#sha256=c81d11edd2b51edf061bdb37a8535885c410b2b7d1a78bb3f40ccf9828dad275 (from https://pypi.org/simple/gpy/)

 Any way this could be fixed?

MartinBubel:
> Unfortunately, I still have to nag about this a bit more. This page shows all available options to install GPy: https://pypi.org/simple/gpy/
> 
> The 1.11.0 version has only one very specific wheel, meaning it won't install on most platforms. Say, it doesn't install on my Ubuntu:
> 
> > Skipping link: none of the wheel's tags (cp35-cp35m-win_amd64) are compatible (run pip debug --verbose to show compatible tags): https://files.pythonhosted.org/packages/89/66/532048790189b5b63c17f5f11532dd4b7aff5a2c4ae9ea02299b4df55cc4/GPy-1.11.0-cp35-cp35m-win_amd64.whl#sha256=c81d11edd2b51edf061bdb37a8535885c410b2b7d1a78bb3f40ccf9828dad275 (from https://pypi.org/simple/gpy/)
> 
> Any way this could be fixed?

I might not be the right guy to answer this but to me it seems like there is no 1.11 release yet (at least there exists no tag). Even worse, the changelog only covers version up to 1.9.8.

Maybe you can limit the upper `GPy` version to `GPy <= 1.10`?

apaleyes:
@MartinBubel there is certainly a 1.11.0 release https://pypi.org/project/GPy/1.11.0/, it's just for one very specific platform. I guess it wasn't stored in GitHub as a release, but that's separate from pip repo.

Issue #1018, opened 2023-06-07
Fix for python3.11?


ego-thales:
Is any release for python3.11 planned?

It is currently no possible to directly install this package under python3.11, and even from sources without diving into the code.
The big package [scikit-fda](https://github.com/GAA-UAM/scikit-fda) depends on it, and so accnot be installed on python3.11 :(

Thank you in advance.

ekalosak:
https://github.com/SheffieldML/GPy/pull/1020 fixes, closing. As the fix is merged, it'll be part of 1.13.0 when released. Please re-open the issue if you run into the problem again after the release.

ego-thales:
Thanks!
Tested installation from latest `devel` rev, it works fine for me :)

Issue #998, opened 2022-11-22
Error on install_longintrepr.h


manueldeprada:
I came across this error while trying to install on Python 3.11 or Python 3.10:

```
gcc -Wsign-compare -DDYNAMIC_ANNOTATIONS_ENABLED=1 -DNDEBUG -O2 -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fstack-protector-strong -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -D_GNU_SOURCE -fPIC -fwrapv -O2 -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fstack-protector-strong -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -D_GNU_SOURCE -fPIC -fwrapv -O2 -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fstack-protector-strong -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -D_GNU_SOURCE -fPIC -fwrapv -fPIC -I/home/prada/.local/lib/python3.11/site-packages/numpy/core/include -I. -I/usr/include/python3.11 -c GPy/kern/src/stationary_cython.c -o build/temp.linux-x86_64-cpython-311/GPy/kern/src/stationary_cython.o -fopenmp -O3
      GPy/kern/src/stationary_cython.c:196:12: error fatal: longintrepr.h: File does not exist
        196 |   #include "longintrepr.h"
            |            ^~~~~~~~~~~~~~~
```

Solution I found: download sources from git, link like ` ln -s /usr/include/python3.10/cpython/longintrepr.h longintrepr.h` and build GPy. 

It still doesn't build in Python 3.11 due API changes like this one:
```
GPy/kern/src/stationary_cython.c:318:11: error: no enough arguments for function ‘PyCode_New’
  318 |           PyCode_New(a, k, l, s, f, code, c, n, v, fv, cell, fn, name, fline, lnos)
[...]

PR #1020, opened 2023-06-15
Rerun with cython 0.29.35 to fix Python 3.11 compatibility issues

jeremysanders:
This should fix the Python 3.11 compatibility problems by rerunning cython using 0.29.35.

See bugs:
https://github.com/SheffieldML/GPy/issues/1018
https://github.com/SheffieldML/GPy/issues/998


bpkroth:
@zhenwendai @MartinBubel, you're some of the last commit authors in the history.  Any chance we can get this committed and a new version cut soon?  There are several projects that rely on it that are currently having issues with recently Python and numpy version mismatches.

MartinBubel:
Thanks for picking this up @jeremysanders . @bpkroth I don't have the rights to merge and even though being among the recent contributors, my experience with `GPy` is very limited.

However, I spend some time with this and did the following:

1. Clone the pull request
2. Run a python3.11-slim-buster container
3. Install python3-dev and build-essential
4. Run `python setup.py develop`
5. Run tests

Running `pytest`, I get one failed test with the following error message:
```python
root@ad1ad64679d9:/GPy# pytest .
================================================================== test session starts ==================================================================
platform linux -- Python 3.11.4, pytest-7.4.0, pluggy-1.2.0
rootdir: /GPy
collected 0 items / 1 error                                                                                                                             

======================================================================== ERRORS =========================================================================
______________________________________________________ ERROR collecting GPy/testing/linalg_test.py ______________________________________________________
ImportError while importing test module '/GPy/GPy/testing/linalg_test.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/local/lib/python3.11/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
GPy/__init__.py:32: in <module>
    from numpy.testing import Tester
E   ImportError: cannot import name 'Tester' from 'numpy.testing' (/usr/local/lib/python3.11/site-packages/numpy/testing/__init__.py)
-------------------------------------------------------------------- Captured stdout --------------------------------------------------------------------
warning in stationary: failed to import cython module: falling back to numpy
warning in coregionalize: failed to import cython module: falling back to numpy
warning in choleskies: failed to import cython module: falling back to numpy
================================================================ short test summary info ================================================================
ERROR GPy/testing/linalg_test.py
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
=================================================================== 1 error in 1.55s ====================================================================
```

I could not immediately figure out what the actual cause of the error is but I would not surprised if this is another "old thing" causing problems nowadays.

@jeremysanders what was your approach on testing?

Anyways "lgtm". But I think it would be better if one the maintainers can take a look at this.

bpkroth:
I took a brief look at this:
> E   ImportError: cannot import name 'Tester' from 'numpy.testing'

That's basically a shim that numpy was providing for running tests with "nose", which is what GPy still uses.

nose is also very out of date and not maintained.  I imagine some hacks could be done to support it, but it's more than I have time to dig into atm.

MartinBubel:
I can take a look at the tests, but not guarantee that I will get it done in the next couple of days... 

But imho it would be best to consider this an opportunity and get rid of outdated / deprecated testing envionrments, so my intended fix would be to update the tests (if possible). Doest that sound good for you @bpkroth ?

MartinBubel:
Also, I wonder whether we should start a new PR for adding some CI job that builds the cython files for each deploy/release.

I don't want to sound too ambitious on this but keeping cython-files in the repo might not be the best idea (as we have seen in the issues leading to this PR).

MartinBubel:
I propse that we merge this PR as this point. As mentioned above, I don't like the fact to much that the cython files are in the repo but that seems to be the case for `GPy` at the moment. And since no github-actions are failing, there is nothing holding us back from merging this.

Regarding the test and deploy changes proposed above, I recommend that we start a separate PR to account for them.

haampie:
Alternatively, remove the generated C sources. That's what cython people advise now, since they cannot guarantee forward compatibility with python.

ekalosak:
Ok, so this one is indeed a bigger fish than the contributor is frying. Thanks very much everyone for bringing this up, and to the contributor for providing an interim fix.

@lawrennd what are your thoughts on the urgency and strategy for (a) removing the generated C (b) digging out of the `nose` tech tebt and (c) merging this PR to support 3.11 in the meantime?

Unless you've got strong feelings on (c) I think it's better to make the pragmatic decision to support 3.11 as soon as possible. For (a) and (b), I suspect there may be less-than-obvious ramifications.

ekalosak:
My one concern with dropping the generated C code is that it may make `pip install GPy` take ages.
@haampie can you link to more info on the guidance for generated C? This _could_ be sticky, best to be sure before we just drop it IMHO.

lawrennd:
I'm less familiar with the technical challenges, but I agree with your analysis. It feels like (c) is a priority, and (a) and (b) are also important but may require some thought. I feel a bit too distant from the technical details to be doing the merge, but I've just sent @MartinBubel an invite on that front.

haampie:
@ekalosak the discussion here: https://github.com/cython/cython/issues/5089#issuecomment-1408714455

ekalosak:
Ok so it seems like the right path forward is to (a) merge this immediately (b) open a discussion issue for removing the compiled cython from the repo.

I'll move my concerns about degrading user build times to that discussion thread.

PR #1031, opened 2023-09-22
cython: list in setup_requires instead of install_requires

haampie:
As far as I can see, it's not a runtime dependency.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -117,7 +117,7 @@ try:
 except ModuleNotFoundError:
     ext_mods = []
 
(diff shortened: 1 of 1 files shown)


MartinBubel (commented):
Hi @haampie 
thank you for the contribution. LGTM.
However, we are looking to remove the precompiled c files from the builds such that they are build when `pip install`'ing the lib.
I have not tested it yet but to my understanding, that would require to keep cython as an install requirement.
Would you agree?

This is why I am not sure if we should merge this, being only that change, as it may change again before the next release.

PR #1040, opened 2023-10-18
1030 move from nose to pytests

### LINKED ISSUE #1030: Move from nose to pytests
It would be great as a community effort if we could move the testing away from nose to pytests. If people are interested in getting involved (which we always welcome!) this could be a good way to start making contributions.

MartinBubel:
Big big PR ...
Did not expect this one to become this big when I started but there was more to migrate than I was hoping for.

A little overview here:
+ All tests, execpt for those in *old_tests* have been migrated to `pytest` which means that `nose` and `unittest` are no longer part of the package.
+ Tests have been renamend and put into the `pytest` convention
+ There also are some "*format on save*" commits which shouldn't be part of this PR but they are ... kind of couldn't resist to doing some formatting here and there.
+ I added `extras: dev` to the *setup.py* file which includes `pytest`, `matplotlib`, and `pods`, which are required for running the tests. It can be installed using `pip install -e .[dev]`
+ I fix some usage of `np.float` and `np.int`. That might introduce some fixes to related issues.

Currently, there are some tests failing:
+ Tests within *test_model.py* are failing because dependency `paramz` uses `np.float`. This is a **problem**! We are planning to get rid of the deprecated types but `paramz` still uses them, `paramz` seems to be "dead" from a maintenance point of view. Does anybody here know about [paramz](https://github.com/sods/paramz/issues/36) and is a developer/maintainer? If we don't get updates there in a reasonable time, we may have to start looking for a replacement.
+ Some tests within *test_mpi.py* seem to be broken. I could not figure out why they are broken now but it does not seem to be related to the `pytest` migration.
+ Some tests within *test_examples.py* seem to be broken. I could not figure out why they are broken now but it does not seem to be related to the `pytest` migration.

Some other things I would like to mention and maybe we can discuss about them in this issue or another:
+ Among the tests (and there are plenty of them), there are many ones that are skipped (maybe too many?)
+ Also, I am not sure if we should keep the *old_tests* in the repo for long as they obviously are *old*  and not used anymore.
+ Also, the amount of *warnings* that are thrown during testing is "significant"

I think we don't need to care too much about these right now but maybe in the long term...

**Open**
+ Some changes are necessary to get the CI jobs running with the new setup (at least I guess). But I thought that its best to start the PR now.
+ Changelog is to be updated

Not sure if someone wants to do a proper review on this PR as there problably are too many changes to track it properly.
My recommendation would be to deactivate those tests that are still failing and take a proper look at them for the next major version update.
Also, the `paramz` issue keeps me worried ... maybe we start a separate discussion on this.


MartinBubel:
@lawrennd @ekalosak any thoughts/objections on this one?

lawrennd:
Great work Martin 

The paramz question is a key one, @ekalosak thoughts on how much effort to remove and utility lost if it's taken out?

MartinBubel:
> Great work Martin
> 
> The paramz question is a key one, @ekalosak thoughts on how much effort to remove and utility lost if it's taken out?

No update here so far but I feel like we should "push" a bit more here as the compatibility issues are rising weekly (even though I am closing some on duplicate).

Due to the "larger than expected" changes necessary in this PR, it is blocking fixes of other issues. Thus I want to get this one merged asap.

@lawrennd I would like to move forward by forking paramz and resolving it for myself, without having to wait for the maintainers at `paramz` to merge. If merged, we can switch back to a "official" paramz release, ultimately but for now, lets use "our own".
Is that fine for you?

MartinBubel:
@lawrennd @ekalosak lets try to keep the momentum. Any thoughts on the above?

I feel like if we don't release a hot fix until the end of the year, it will be even more difficult to get the required maintenance done ...

lawrennd:
Thanks for the bump. Yes, I think forking is the best idea.

MartinBubel:
To me, this one is ready now.

Changed testing from deprecated combination of `unittest` and `nose` to `pytest`.
On that route, there are two tests that are failing for reasons I could not work out. Thus, I deactivated those tests.

I propose to merge this one so we can get ahead on our path to hot-fixing the lib to provide better support for modern python setups.
We can open separate issues for reactivating the two tests.
Also, I would consider ignoring the codecov checks for this MR (as its way to big and still mandatory).

@lawrennd are you fine with me merging this one and continuing with the maintenance-issues?

lawrennd (approved):
Thanks for this Martin!

lawrennd:
Yep, happy for you to merge, thanks Martin.

PR #1045, opened 2023-11-16
limit minimum python version #1038

### LINKED ISSUE #1038: Limit maximum python version
We currently don't put a limit on the maximum python version supported - at least to my best knowledge.

#1036 raised this for a good reason and was only closed for practical reasons.

The next release should put a limit the python version to 3.11 at max.

MartinBubel:
We decided on supporting the same minimum version as numpy. Thus, we move to minimum python3.9


MartinBubel:
Hi @lawrennd 
sorry for the high traffic (there might be more coming up).

There is not too much in this MR. Simply a limitation to the minimum supported python version to prevent further compatibility issues with modern numpy versions.

This may drop support for some users but python 3.8 is end of life, so I guess moving to a minimum python 3.9 is "fair".

PR #1048, opened 2023-11-24
Fix deprecated function readfp

me2d09:
Since Python 3.2 function readfp is replaced with read_file: [python documentation](https://docs.python.org/3.12/library/configparser.html#configparser.ConfigParser.read_file) 

Readfp was fully removed in Python 3.12 so this fix is needed for GPy.


MartinBubel:
Hi @me2d09 
thank you for your work!

PR #1065, opened 2024-02-03
Correct dl_dm term in student t inference

tangoed2whiskey:
The dl_dm term in the gradient of the log-likelihood for the Student T process regressor seems to have been copied directly across from the Gaussian process regressor: however it should be slightly different: correction included here.

(My formatter has also adjusted a couple of other lines that are not functionality changes)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Correct dl_dm term in student t inference #1065
+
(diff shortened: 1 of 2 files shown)


MartinBubel:
Hi @tangoed2whiskey 
thanks for the contribution. Took me a while to find the time to check this.
I'll update the changelog and merge it.

MartinBubel (approved):
lgtm!

PR #1080, opened 2024-07-15
Switch to poetry

MartinBubel:
The issues regarding dependency restrictions and pep-related updates are stacking so I decided it may be time to continue with this act of modernization.


MartinBubel:
Hi @lawrennd ,
I have prepared this PR to migrate to python-poetry, which would allow for a better dependency tracking and communication (considering the increasing number of issues related to numpy and maptlotlib compatibility).

However, it seems like the combination of Apple + M-Series chip still fails (CI tests). Unfortunately, debugging on CI runners is tedious. I've gotten access to an old macbook with intel chip and I cannot reproduce the issue there so it most likely is related to apples new M-series chips.

For me, there are three options and I would like to get your opinion on them:
1. Drop macos support for now. This is not a nice thing to do for end-users, though I would suspect that academic and industrial users might be able to live with it.
2. Find some cloud service that provides apple silicon runners, like [hello-m1](https://www.scaleway.com/en/hello-m1/) or wait until a friend borrows me a mac.
3. Stick with setup.py for now.

PR #1103, opened 2024-10-27
add ruff pre-commit

### LINKED ISSUE #1102: add ruff linter and formatter

MartinBubel:
Comes at a "problem": ruff identifies almost 800 errors than cannot be fixed automatically and require user interaction.

This is quite tedious and I cannot address it atm as it is just too much work.
Running ruff with `--unsafe-fixes` brings that number down to just 255 files, but I do not feel comfortable with simply applying those potentially unsafe operations as I am not sure whether the tesitng lib covers all potential issues.

@janmayer any idea / input / experience on this?

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,12 @@
+repos:
+- repo: https://github.com/astral-sh/ruff-pre-commit
+  # Ruff version.
+  rev: v0.7.1
+  hooks:
+    # Run the linter.
+    - id: ruff
+      types_or: [ python, pyi ]
+      args: [ --fix ]
+    # Run the formatter.
+    - id: ruff-format
+      types_or: [ python, pyi ]


janmayer:
The only thing you can do here is go slow. 

First, use `ruff format` (maybe with some line length config) and ensure that all further commits follow the formatting (via a github action). That will create one massive commit, but as everything is just a format change, it will be save. This will prevent a lot of commit noise in the future (like the whitespace in my earlier PR).

From there on out, go bit by bit - autofix missing imports and import order only, check if everything still works. Its not necessarily truly "save", e.g. if file A imports things from file B which it has imported from C, but are not used in B.

That will probably reduce the number of warnings extensively already.

janmayer:
For example, start with `pyproject.toml` like so

```
[tool.ruff]
line-length = 120
indent-width = 4
target-version = "py39"

[tool.ruff.lint]
# https://docs.astral.sh/ruff/rules/
select = ["F401", "F404", "E401", "E402", "I001", "I002"]

[tool.ruff.per-file-ignores]
"__init__.py" = ["F401"]  # TODO: Use __all__ in __init__.py
```

Autofixing will solve quite alot of "problems", but not some sub-cases (`F401 `IPython` imported but unused; consider using `importlib.util.find_spec` to test for availability` to replace the try import blocks, and `E402 Module level import not at top of file` when it can't tell if its save to move). If you don't want to change these manually yet, you can comment out  `E402` and `F401` afterwards.

MartinBubel:
Thanks for your input on this. I appreciate it!
I'll be back on this sometime later this week.

PR #1108, opened 2025-01-04
fix: pickle and deep copy classification models with EP

### LINKED ISSUE #1105: Errors when copying or pickling models using the expectation_propagation.EP() class
## Problem

When I try to deep copy or pickle a binary classification model implementing the `EP()` inference method (from the `GPy.inference.latent_function_inference.expectation_propagation` module), I get the following error:
```
  File "[...]/site-packages/GPy/inference/latent_function_inference/expectation_propagation.py", line 238, in __getstate__
    return [super(EPBase, self).__getstate__() , [self.epsilon, self.eta, self.delta]]
AttributeError: 'super' object has no attribute '__getstate__'
```

## Reproducible example

Ran with Python 3.9 and GPy version 1.13.1:

```python
import GPy
import copy
import numpy as np
import pickle as pkl

# Dummy binary classification dataset
X = np.array([0, 1, 2, 3]).reshape(-1, 1)
Y = np.array([0, 0, 1, 1]).reshape(-1, 1)

# Model
m = GPy.core.GP(
    X=X,
    Y=Y, 
    kernel=GPy.kern.RBF(input_dim=1, variance=1.0, lengthscale=1.0), 
    inference_method = GPy.inference.latent_function_inference.expectation_propagation.EP(),
    likelihood=GPy.likelihoods.Bernoulli(),
    mean_function=None
)

copy.deepcopy(m)  # triggers error
pkl.dumps(m)  # also triggers same error
```

## Suggested fix

I believe the issue is that the parent class [`EPBase`](https://github.com/SheffieldML/GPy/blob/9a318860856e54990257c5e49c7e864585ffae8d/GPy/inference/latent_function_inference/expectation_propagation.py#L187) only inherits the basic `object` class. Calling `super(EPBase, self).__getstate__()` will fail because the `object` class does not implement this method. In fact, some other method calls with `super()` in the `EPBase` class should also fail (such as `super(EPBase, self).__setstate__(state[0])` and `super(EPBase, self)._save_to_input_dict()`).

As a fix, I think these `super()` method calls can simply be removed in the `EPBase` class.

olamarre:
- Removed unecessary code preventing pickling & copying
- Wrote a new test to check that pickling and copying works

Fixes #1105

Longer description:
- Removed all the "super" commands in the `EPBase` class since it does not have parent classes (raises an error otherwise - I suspect methods employing this command were never used or tested).
- Removed the `__setstate__` and `__getstate__` methods as they did not fully capture the state of an instance, making the implementation brittle. An error would be raised when pickling and unpickling an instance (for example, the `self.always_reset` attribute was missing). By removing these instances, Python now defaults to using the `__dict__` method of the class, which works well and is much simpler.
- Added a test to validate that picking and deep copying now works with a classification model using expectation-propagation (ran test with pytest 8.3.4, GPy version 1.13.2 and Python 3.9.19)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Changelog
 
 ## Unreleased
+* fix pickle and deep copy for classification models inheriting from EP #1108 [olamarre]
+
 * update prior `__new__` methods #1098 [MartinBubel]
 
 * fix invalid escape sequence #1011 [janmayer]
--- a/GPy/inference/latent_function_inference/expectation_propagation.py
+++ b/GPy/inference/latent_function_inference/expectation_propagation.py
@@ -229,24 +229,17 @@ class EPBase(object):
         v_diff = np.mean(np.square(ga_approx.v-self.ga_approx_old.v))
         return ((tau_diff < self.epsilon) and (v_diff < self.epsilon))
 
-    def __setstate__(self, state):
-        super(EPBase, self).__setstate__(state[0])
-        self.epsilon, self.eta, self.delta = state[1]
-        self.reset()
-
-    def __getstate__(self):
-        return [super(EPBase, self).__getstate__() , [self.epsilon, self.eta, self.delta]]
-
     def _save_to_input_dict(self):
-        input_dict = super(EPBase, self)._save_to_input_dict()
-        input_dict["epsilon"]=self.epsilon
-        input_dict["eta"]=self.eta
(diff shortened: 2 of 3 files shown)


MartinBubel:
thanks @olamarre for your contribution. LGTM!
I have updated the changelog (noting your contribution) and I will merge it after the pipeline finishes.

If you need a new release featuring the fix, raise your hand - then I'll trigger the release pipeline.

PR #1115, opened 2025-06-12
Fix lint issues related to comparisons with None

yurivict:


MartinBubel:
I assume that tests are failing because of pytest>=8.0.
I‘ll update the workflow and see if it succeeds then.

yurivict:
These are obviously legit fixes.
You can just merge this PR without waiting for tests.

MartinBubel:
Sure, no doubt about that.
But I won‘t merge a broken CI into devel.

I‘ll fix the CI in another issue and rebase your PR.

